Issue #1, opened 2021-06-28
Main is a script => make it an executable module


Didou09:
Motivations:
----------------
* scripts are executed at import time
* Separating import from execution os a good practice 
* If converted to a module it would allow to include other functionalities the main one
* It can still be an executable `if__name__ == "__main__"`

Proposition:
----------------
* Convert to a module, with a run() function and a `if__name__ == "__main__"` at the end to make it executable

PR #5, opened 2021-06-28
Issue001 main as exec module

Didou09:
Motivations:
----------------
* An executable module:
    - still allows to execute the code in a single terminal line
    - Allows the user to pass input argument directly from the command line instead of having to hard-code then by editing the file
    - Allows for a cleaner code and more flexibility
    - Separates import from execution, which is also useful for advanced debugging and flexibility

Proposed main changes:
----------------------------------
* In Main.py, there is now:
    - a _**shebang**_ at the top of the file
    - a `Main.run()` function that is callable from python and takes arguments and does the same thing as the former script
    - Several utility hidden functions to handle input arguments (fro python and from terminal)
    - A `if __name__ == '__main__` section at the end to handle argument parsing from the terminal using library `argparse`
* An extra unit test was added to make sure executing the module from the terminal was working properly
* Minor refactoring was performed (spaces after commas, indentation, lines length, ordering of imports...)

Issues:
----------
* Addresses issue #1 

Example:
-------------
* First make sure you `Main.py `file is executable (`chmod +x Main.py` in linux)
* The execute the following line from inside the project's folder:
`./Main.py --plot False --timeit True`

(depending on your platform the exact command to use may vary a bit)

### DIFF
--- a/Main.py
+++ b/Main.py
@@ -45,6 +45,7 @@ WHAT I AM (Paul) LOOKING FOR IN FURTHER DEVELOPMENT
 # Built-in
 import os
 import time                     # Time (run speed) printing
+import argparse
 
 
 # Library-specific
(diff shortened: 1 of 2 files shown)


DaluS (approved):
I trust you on this :D

PR #18, opened 2021-07-01
Parameters + Gregoire protomodel

DaluS:
+Readme (partial)
+Parameters comments / units
+Gregoire's thermodynamic ressource
+Correction of a few tweaks


DaluS:
So the error is :

image

If I look at the code 

image

I don't understand the error. 
The code run on my computer

Didou09 (commented):
Thanks for all this work,

If I may I have several comments:
* It is a bit dangerous to put all those independent changes in a unique pull request
* Better practice would have been to do a separate Pull Request for each independent change, from what I understand you already identified the following independent changes
    - Adding the  README
    - Adding units / comments to the parameter dictionnary
    - Adding Gregoire's toy model
    - Minor changes / tweaks
* Do you think you could find time to make a separate branch and pull request for each ? (create a branch from devel for each, separately)

The reasons behind this approach would be:
* Too many changes at once can be difficult to follow for reviewers (sometimes there is no choice, but here it's possible to split)
* When there is a bug and tests are not passing (like in this case) it's more difficult to find the bug when there are many independent changes
* SInce tests are not passing, the PR cannot be merged as such, which means many independent changes that probably did not cause the bug cannot be merged and used in devel yet (adding the README and completing units probably did not cause the bug, we cannot merge these now because they are in the same PR as the changes that caused the bug, if they had been in their own PR they could have been merged already).
* Good practice is one issue = one PR, here we have 4 issues in one PR

I know it can be frustrating when all the work is done on one PR, but it's definitely better practice and in the current situation we cannot merge. Seperating the 4 issues will make things easier, if you have the time.

As an example I feel the units / dimensions of parameters need to be discussed and clarified further, some of them probably would deserve to be discussed in more details, a dedicated PR would be the right place for this discussion (see some comments on this below).  And others changes like adding the README could be merged without having to wait for that discussion to be solved (or for the bug to be found)

_def_parameters.py
```diff
@@ -267,25 +255,25 @@
     'tauR': {
         'value': 2.0,
         'com': 'Typical time for recruitement',
-        'units': 'y',
+        'units': 't',
```
Didou09, line 258: I might be misunderstanding, but I feel in all those cases the units really is 'y', and 't' is the physical dimension.
If tauR = 1, the user will probably want to know if it's 1 year or 1 month I guess ?

A way to circumvent this would be to include the 2 keys in the dict for each parameter:

    'tauR': {
        'value': 2.0,
        'com': 'Typical time for recruitement',
        'units': 'y',
        'dimension': 't',        # or 'time' ? as you prefer
        'group': 'RelaxBuffer',
    },

What do you think?

_def_variables.py
```diff
@@ -12,49 +12,138 @@
...
             'value': None,
-            'com': '',
+            'com': 'employement rate',
             'units': '',
         },
         'd': {
             'value': None,
-            'com': '',
+            'com': 'relative private debt',
             'units': '',
```
Didou09, line 53: units = 'dollars'            ?
dimension = 'money'   ?

_def_variables.py
```diff
@@ -12,49 +12,138 @@
...
-            'units': '',
+            'com': 'Capital',
+            'units': 'Capital',
         },
         'W': {
             'value': None,
-            'com': '',
-            'units': '',
+            'com': 'Salary',
+            'units': 'dollars',
```
Didou09, line 31: dimension : 'money / wooker / time'    ?
units: 'dollars / worker / y'  ?

_def_variables.py
```diff
@@ -12,49 +12,138 @@
...
-            'units': '',
+            'com': 'productivity per worker',
+            'units': 'Output per person',
         },
         'N': {
             'value': None,
-            'com': '',
-            'units': '',
+            'com': 'Population',
+            'units': 'pop',
```
Didou09, line 21: dimension: 'humans'             ?
units: 'humans'                      ?

_def_variables.py
```diff
@@ -12,49 +12,138 @@
     'GK': {
         'a': {
             'value': None,
-            'com': '',
-            'units': '',
+            'com': 'productivity per worker',
+            'units': 'Output per person',
```
Didou09, line 16: dimension: 'output / person'    ?
units:   'dollars /person '   ? (is output in monetary or physical units ?

_def_variables.py
```diff
@@ -12,49 +12,138 @@
...
+            'com'  : 'Exogeneous radiative forcing',
+            'units': 'W m^{-2}',
+        },
+                
+        ### TYPICAL ECONOMY VALUES
+        'd': {
+            'value': 1.53,
+            'com': 'relative private debt',
+            'units': '',
+        },
```
Didou09, line 105: dimension: 'money'      ?
units: 'dollars'              ?

Or is it dimensionless ?

DaluS: It's an intensive unit, hence dimensionless :)

_def_variables.py
DaluS: This is indeed a propblem, some articles use Y as a physical output and some people use Y as a monetary output. We can create two different variables if we want to be as general as possible ( with Y_{nominal}= Y_{real} p )

_def_variables.py
DaluS: As said in a comment below (I'm going bottom up) It's an intensive unit, hence dimensionless :)

( The definition is output in dollars pY divided by the debt D )

Didou09:
Congrats for finding the bug, I forgot to asnwer an error syaing EOF is usually something related to unclosed perenthesis, dict, list...

_def_parameters.py
DaluS: You are right, BUT it will makes the structure much heavier to use and develop, I feel there is a very slippery slope of bringing more details in the code for years without using it as we allocate all our energy to the development of the code itself and not on the physics and the analysis of what we code. The edge we have to follow is thin on this. 

I'll recheck the quantities and be sure we have consistent units everywhere.

Didou09:
So now the bug is fixed, the tests pass and we could merge, but should we ?
Usually, before merging it is good habit to:
* Answer the questions of all reviewers
* If there are issues raised by these questions, that are solved by this PR, a dedicated issue shall be opened for each, separately

So in this case it would mean:
* Providing an answer to the questions on units / dimensions, and possibly deciding that the PR can be merged provided a separated dedicated issue s opened for this.
* Deciding that the PR can be merge as such (i.e. without splitting into 4 smaller PR), although splitting would be better, at least nothing is broken and we can try to remember this advice for next time

Your choice ;-)

Didou09 (approved):
Bug is fixed, as a reviewer I am approving the PR provided the questions raised are answered here or a dedicated issue is created, and provided we try to have seperate PR for independant issues next time

Congrats for your first reviewed and validated PR !

_def_variables.py
DaluS: Corrected !

Didou09:
Other piece of advice: If the PR is still work in progress, you can turn it into a draft to tell everyone it's not finished:
image

DaluS:
I found the mistake indeed ! I think we can merge. I also answered to all the points. 

There is one unit I think I'm going to change (again) as I separated in units workers and persons, I will put human instead of population as the general unit for each

DaluS:
> Other piece of advice: If the PR is still work in progress, you can turn it into a draft to tell everyone it's not finished:
> image

I didn't know ! Sorry for the mess, I'll definitely do it next time :)

_def_parameters.py
Didou09: Yes, that's why it would be nice to have this unit issue in a separate dedicated PR, so we can live without it and just merge it when it's ready, if someone finds time to deal with it.
This way it is contained on a dedicated branch and is only merged if / when it's ready

_def_variables.py
Didou09: Yes, that might become necessary, and indeed Regionalized databases like Exiobase can provide both, so it would make sense (but let's keep this for a future separate branch for cleanliness, and to avoid the temptation of doing everything in the same branch / PR, which quickly becomes a nightmare for tracability)

PR #28, opened 2021-07-07
Intermediate stable PR : functions handled

Didou09:
Disclaimer:
------------
* This is still an intermediate PR, but it is stable and does not break the unit tests

Main changes:
----------------
* `_FLATTEN` was indeed useless and removed
* `_def_fonctions.py` is introduced, where functions are defined
        - I used `def` instead of `lambda` because lambda does not allow for docstrings and is unnamed, which can sometimes make things harder for debugging, please confirm this is ok for you @DaluS 
        - As you suggested, all functions can take parameters and variables, and even other functions as input argument
        - The code reckognizes each type of input argument and displays them in 3 groups (parameter, variable, function) in the `get_summary()` output

...To be finished (bus leaving)

### DIFF
--- a/_class.py
+++ b/_class.py
@@ -8,6 +8,13 @@ import numpy as np
 # Library-specific
 import _utils
 import _class_checks
+import _class_utility
+
+
+# #############################################################################
+# #############################################################################
+#                       Main class
+# #############################################################################
 
 
 class Solver():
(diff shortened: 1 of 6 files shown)


DaluS:
Ok it's very clean but I do not think it is the right approach to keep the code modeler-practical.
It's very hard using this type of representation to understand exactly what is or not in a model, adding elements or modifying them. 
It also needs a lot of data to complete each that are redundant with the definition of variables in `_def_variables`. 

I know it's better for the end-user but we are nowhere near that step, it's so much heavier to work with such structure... We need to talk about the architecture, where we want the user and the modeler to write so that it is as practical as possible for both of them.  For me concision in the code, the ability for one to understand the architecture (what is where) is the most important point.  

Especially a model is a set of functions through hypotheses, and as such the set should be consistent and always fully defined at the same place with the expressions. it's much more practical to verify that way the integrity of the system, and maybe go later into an overlay in the construction of models ( using stock-flow tables )

* **About lambda functions** : there is no docstring indeed but it is defined with a comment inside the system of equation, I think it works well that way ? 
* **About test on the functions** : the values of all the functions can be restored at each step of the system. Indeed I don't use debugging tools as I never had to, using an IDE and ipython approaches always gave me more than enough control to verify that type of thing. Indeed the fact that we put the code inside main with no output make it more complex and I happened to develop by executing lines-by-lines of this code and some other methods should work better. Also, the structure we put into place like this allow each function to be fairly simple (it's at most three variables and parameters ), I don't think this point will be a problem. Finally, I think unitary tests would be fairly easy to add on such simple equations. 
* **Functions in entry of functions** : I don't see exactly the point of doing that ? The order of functions being automatically determined I assume it should be less necessary. 

I think at this point we should discuss about the general philosophy of the code : 
* What exactly is the purpose ? 
    * Allowing rapid development of consistent model ? 
    * Allowing the study of multiple models ? 
    * Allowing the extensive of very complex models with an extensive optimization of run ? 
    * Having a code friendly for the non-coding user ?  For the model developer ? For the code developper ? 
* what is the exact architecture ? 
    * Numerically, how does it flows between the different files ? 
    * In term of development, where should a modeler add file and lines ? 
    * In term of use, where a user should edit files or send arguments ? 

Doing this will (I think) give us a good vision on what we should focus or not our efforts on, how to have a good repartition of task, and what direction we should take in the organization. 
As the PR is not changing the run itself I'll merge it, then I'll try to answer the questions above and define the different development axes ! 
I think it will be necessary to have a meeting after to be sure we are well synchronized on all of these points :) Tell me when you are available, I'm almost free anytime this week !

PR #29, opened 2021-07-08
NEW STRUCTURE, AND MORE CLARITY

DaluS:
Quite a lot of change but it's much easier to have a good overview of the model now. 

## About users

There are three categories of people who are going to work on the subject : 
* The developpers ( for the moment Didier and I )
* The modellers ( Camille, Hugo B, Gregoire )
* End-users ( Mlondi for example )

developpers needs to have a good overview of the whole structure, modellers of everything at the root and the subfolder of models, and End-user only files made for them (`overlay.py` for example and equivalent files). 

## The classification of files

I added folders (sorry for the french I was thinking in french this afternoon) : 
 * Un sous-dossier avec tous les codes utils/check automatique (dans lequel seul le developpeur met le nez)
 * Un sous-dossier avec les tests unitaires 
 * Un sous-dossier avec tous les systemes d'equations qui y sont resolus
 * Un sous-dossier avec les codes non fini d'implementer / deprecie mais conserve car contenant des choses utiles ( qui vont ensuite dans les tests / utils )
 * Un sous_dossier sauvegarde (pas necessairement au meme endroit mais par defaut oui), pas encore code

## The flow

1. The user decides if he wants to use a command-line execution or an Overlay
2. If he uses the overlay, he fills (or execute a part that fills it itself) the kwargs of main
3. Main.py : 
	* **Check**       : check if the input values are correct
	* **load**        : the required module of equations, and list all variables/parameters required 
	* **Prepare**        : all the parameters required, all the variables required
	* **Announce**    : Gives all informations the user should be looking for about this run
	* **Computation** : Solve temporally the system
	* **Expansion**     : Create a nice solution 
	* **Output**          : Do typical plots / Save / Return

## The files 

* **The system of equation** we are going to solve is is in the folder SystemOfEquations
* **Overlay** will allow some fancy computing through main control (sensibility analysis, attraction basin...), and is a layer that control the main (jupyter-friendly use)
*  **Main** is the main loop
* **Class** contain the core method for resolution. 
* **Parameters** Contains all initial values for all existing parameters, the system only take the one needed.
*  **VariableDictionnary** gives information on what are the variables and how to plot them

### Integrate Didier's modifications 

* Didier is working on modification on : 
    * Class (through `_class`, `_class_checks`, `_class_utility`) that should be integrated on necessary functions 
    * Parameters (through `_def_parameters`), which should replace `parameters.py`. I think it's almost ready and should be the next thing. Connexion are in part **prepare** and `__init__` (change `parameters` which is the short version 'key-value' and `ParametersDescription` which is the full version witth only relevant variables) . The print in class should be also modified (they already exist in `_def_parameters`
    * Variables (through `_def_variables`), some elements 

### Some other modifications 

* Calculate all intermediary variables just after initialisation and print them ( so that the user can verify the consistency of initial conditions) 
* Add the save part
* Add all input check (most variable are right now not checked)
* Add an unitary test of model verification ( parameters exist, variables exist, there is an order to solve the system)
* Do the list of unitary tests we should add
* I coded a function that read the python file as text and returned the list of parameters if the modeller is lazy. We can do that if we want ! 
* Discuss about input-allocation-output in the rk4 for fastest resolution
* Add an adaptative timestep based on the computed (dy/y * dt) (we want y/dy * security= dt for each dynamical variable)
* Discuss what are the keys necessary for parameters and functions
* Find an elegant way in the choice of model resolution, import all models in the folder
* Find cool plots template code to do some fancy one !


DaluS:
I have done a last-minute mistake everything was working well on my computer 20 minutes ago. I have done a kind of merge manually with my files which has created chaos... 

I should : 
* only do merges through github
* Do more commit 

I'm still learning on this :) 
As it is 9.30 pm right now I'll clean that tomorrow ! The overall structure is working well if we forget that weird bug ( I think _def_variables inside is outdated also. Will double-clean it on my morning)

Didou09:
Yes, as you have noticed the key for a clean collaborative work on the long term is:
* many _small_ commits
* PR via Github

This way you should be safe from chaos :-)

DaluS:
Alright the code was perfectly running BUT in the folder `NotIncludedYet` there was some code not finished with errors, I hadn't thought about this. 

This release is stable ! Overlay not getting through the argparser correctly for the moment but everything else is working

PR #46, opened 2021-07-09
Issue006 choice of system of equations

Didou09:
Disclaimer:
---------------
* This is work in progress, but is now stable, and quite complete, I think we are finally getting there :-)
* It is fully operational, you use it (see tuto below) and if you agree and merge, we can now work from that
* This is hopefully my last big PR, and if you agree, from there we can work with small PR, small branches, small changes
* There are still a few minor changes to makes: a bit of code cleaning and fixing PEP8 issues (I'll do it tomorrow)

Main changes:
-------------------
* **Robustness vs developer**: 
    - At import time, a check is run on all fields of all default parameters and identifies any field not matching a known value (there is list of allowed units, dimensions...)
    - This is temporary, and can be used as a help for completing missing fields, finding typos...
    - If any non-conformity is found, a warning message listing them is printed
    - @Dalus, you're probably the most relevant person to double-check these fields
* **Robustness vs user**: 
    - A lot more tests have been implemented to check inputs (parameters and functions)
    - Detailed error messages have been prepared for most common mistakes
* **Function order**:
    - If the user provides an order, its conformity is checked
    - If the user does not provide one, the code find one and sets it (and prints a message showing the various possibilities)
* **model loading**:
    - Users who want to add models have to create a `_model_NAME.py` file in models/
    - There is no need for the user to add it to `__init__.py` like before, this is now transparent and managed automatically
    - There is now a function to list all available models (see below)
    - Model loading can now be done at instanciation
* **unit tests**:
    - A new unit tests file has been added to run tests on the `Solver` class
* **Introspection**:
    - The `__repr__ `method is rewritten to print a mini-summary when the user just types the instance
    - The `Solver.get_summary()` method is updated to include the detail of each function
    - The `Solver.get_dparam()` method is updated to allow selection by `eqtype`
* **Simulation info**:
   - Like you implemented, there is an optional info printing during the run
   - It is set by argument verb, if verb=1 a dynamic progress is printed (default), otherwise (verb=2) it's one line per time step 
* **output**:
   - All results are stored in a variable for each function
   - But the user can ask for a compacted numpy array containing all variables, if needed

Typical usage:
-------------------
```
In [1]: import _class

In [2]: sol = _class.Solver('GK')               # select model from single key
['Y', 'L', 'Pi', 'lambda', 'omega', 'philips', 'kappa', 'I']

In [3]: sol.run()                                        # run (with progress bar)
time step 10000 / 10000

In [4]: sol.plot()                                       # plot (to be compleetd by you ?, method does nothing right now)

In [5]: sol.save()                                     # save (to be completed by me, cf. issue #3 )
```

Detailed Examples:
--------------------------

Warning against non-conformity at import time (temporary, for developers)
```
In [1]: import _class
/home/didier/Documents/Projects/GEMMES-1/models/_def_fields.py:946: UserWarning: The following keys of _DFIELDS are non-conform:
	- time: ['Time']
	- muI: ['NOTDONEYET']
	- muN: ['NOTDONEYET']
	- C: ['SI']
	- C0: ['SI']
	- FexoMax: ['W M^{-2}']
	- a: ['Productivity', 'Output Humans^{-1}']
	- N: ['10^{9} Humans']
	- K: ['Technological Unit']
	- CO2at: ['GtC']
	- CO2up: ['GtC']
	- CO2lo: ['GtC']
	- E_ind: ['GtC Y^{-1}']
	- E_land: ['GtC Y^{-1}']
	- F_exo: ['W m^{-2}']
	- Yn: ['10^{12} Dollars']
	- g_sigma: ['Y^{-1}']
	- Pi: ['Money?']
	- g: ['?']
	- Y: ['Missing']
	- L: ['Missing']
	- i: ['Missing']
	- I: ['Missing']
  warnings.warn(msg)
```

Get list of available models (more details can be added later) + a more direct function can be implemented
```
In [5]: _class._class_checks.models.get_available_models()
The following predefined models are currently available:
	- GK
	- GK_Reduced
```

Instanciating and loading model at the same time + printed the automatically suggested functions order
```
In [2]: sol = _class.Solver('GK')
['a', 'N', 'time', 'Y', 'L', 'Pi', 'lambda', 'omega', 'philips', 'kappa', 'W', 'I', 'K', 'D']

```
@Dalus: question: `philips` vs `phillips` ?

A parameter value can be changed using: 
```
Solver.set_dparam(key='k2', value=0.15)
```

Printing a very short summary of the model
```
In [3]: sol
Out[3]: 
model  source                                                        nb. model param  nb. functions  run  
-----  ------------------------------------------------------------  ---------------  -------------  -----
GK     /home/didier/Documents/Projects/GEMMES-1/models/_model_GK.py  19               18             False
```

Printing detailed summary
```
In [4]: sol.get_summary()
Numerical param.  value  units  comment                                                   
----------------  -----  -----  ----------------------------------------------------------
Tmax              100    y      Duration of simulation                                    
nx                1      None   Number of similar systems evolving in parrallel           
dt                0.01   y      Time step (fixed timestep method)                         
Tstore            0.01   None   Time between storages (if StorageMode=full, it goes to dt)
nt                10000  None   Number of temporal iteration                              
ns                None   None   Number of elements stored                                 
storage           full   None   flag indicating which time steps to store                 
save              True   None   flag indicating whether to save output data               

Model param.  value                  units   group       comment                                                    
------------  ---------------------  ------  ----------  -----------------------------------------------------------
r             0.03                   y^{-1}  Prices      Interest at the bank                                       
k0            -0.0065                None    Keen                                                                   
k1            0.006737946999085467   None    Keen                                                                   
k2            20                     None    Keen                                                                   
alpha         0.02                   y^{-1}  Population  Rate of productivity increase                              
delta         0.005                  y^{-1}  Capital     Rate of capital depletion                                  
beta          0.025                  y^{-1}  Population  Rate of population growth                                  
nu            3                      None    Production  Kapital to output ratio                                    
phinull       0.04                   None    Philips     Unemployment rate that stops salary increase (no inflation)
phi0          0.04006410256410257    None    Philips                                                                
phi1          6.410256410256412e-05  None    Philips                                                                

function  source                       shape       units               eqtype        comment                                           
--------  ---------------------------  ----------  ------------------  ------------  --------------------------------------------------
a          alpha * itself              (10000, 1)  Output Humans^{-1}  ode           Exogenous technical progress as an exponential    
N          beta*itself                 (10000, 1)  10^{9} Humans       ode           Exogenous population as an exponential            
K          I - itself*delta            (10000, 1)  Technological Unit  ode           Capital evolution from investment and depreciation
W          itself * philips            (10000, 1)  Dollars             ode           Wage evolution through philips curve              
D          I - Pi                      (10000, 1)  Dollars             ode           Debt as Investment-Profit difference              
Y          K / nu                      (10000, 1)                      intermediary                                                    
L          K / (a * nu)                (10000, 1)                      intermediary                                                    
Pi         Y - W*L - r*D               (10000, 1)                      intermediary  Absolute profit?                                  
lambda     L / N                       (10000, 1)                      intermediary  employment rate                                   
omega      W * L / Y                   (10000, 1)                      intermediary  Wage share of the economy                         
philips    -phi0 + phi1 / (1-lamb)**2  (10000, 1)                      intermediary                                                    
kappa      k0 + k1 * np.exp(k2*Pi/Y)   (10000, 1)                      intermediary                                                    
I          Y * kappa                   (10000, 1)                      intermediary                                                    
g          (1-omega) / nu - delta      (10000, 1)                      auxiliary                                                       
d          D / Y                       (10000, 1)                      auxiliary     relative private debt                             
pi         1 - omega - r*d             (10000, 1)                      auxiliary     relative profit?                                  
i          Y*0                         (10000, 1)                      auxiliary                                                       
time       dt                          (10000, 1)  y                   ode           Time vector      
```

Print only a category of parameter / functions
```
In [5]: sol.get_dparam(eqtype='ode', returnas=False)
parameter  value       units               dim.          symbol  type       group       comment                                           
---------  ----------  ------------------  ------------  ------  ---------  ----------  --------------------------------------------------
a          (10000, 1)  Output Humans^{-1}  Productivity  $a$     intensive  Economy     Exogenous technical progress as an exponential    
N          (10000, 1)  10^{9} Humans       Humans        $N$     extensive  Population  Exogenous population as an exponential            
K          (10000, 1)  Technological Unit  Money         $K$     extensive  Economy     Capital evolution from investment and depreciation
W          (10000, 1)  Dollars             Money         $W$     extensive  Economy     Wage evolution through philips curve              
D          (10000, 1)  Dollars             Money         $D$     extensive  Economy     Debt as Investment-Profit difference              
time       (10000, 1)  y                   time          $t$     extensive  Time        Time vector    

In [6]: sol.get_dparam(units='y^{-1}', returnas=False)
parameter  value  units   dim.       symbol    type       group       comment                      
---------  -----  ------  ---------  --------  ---------  ----------  -----------------------------
r          0.03   y^{-1}  time rate  None      intensive  Prices      Interest at the bank         
alpha      0.02   y^{-1}  time rate  $alpha$   intensive  Population  Rate of productivity increase
delta      0.005  y^{-1}  time rate  $\delta$  intensive  Capital     Rate of capital depletion    
beta       0.025  y^{-1}  time rate  $beta$    intensive  Population  Rate of population growth    

```

Run the simulation
```
In [10]: sol.run()
time step 10000 / 10000
```

All variable are stored in the same dict as parameters and functions
```
In [11]: sol.get_dparam()['K']['value']
Out[11]: 
array([[0.1       ],
       [0.09999322],
       [0.09998645],
       ...,
       [0.23048922],
       [0.23047274],
       [0.23045627]])
```

All variables can be obtained in a single compacted array, if needed:
```
In [4]: lkeys, array = sol.get_variables_compact()
```

All variables can be reset (although it's done automatically before running)
```
In [5]: sol.reset()
```

Next steps:
---------------
* If  you agree, you can merge (after I fix the PEP8 issues)
* If you want to see some major changes implemented, don't merge, list them and I'll try to work on them
* If you are ok and merge, we will implement all necessary small changes after the merge, using small well-defined issues and dedicated branches
* Everything is ready for the data saving, which I have prepared a method for, and which I will take care of after the merge
* I have also prepared a plot() method that you can use to implement the plotting methods you have already started implemented 
* Unit tests are running and passing, more will come as we continue developping 
* There are many nice things to do, among the ones I can take care of is the conversion of this into a real python package, if you want (packaging, deployment...)


DaluS:
> * I think the function order is part of the model, because two models with the same parameters and functions but with different function orders may give numerically different results
>   => I suggest we have a function that suggest a clever order (being implemented) but that we always leave th epossibility for the user to force the order he / she finds most relevant
>   => Of course, when the class will allow for saving output, the functions order should be saved too, for reproducibility

* The first point should not be true : all working order are giving the same system (because intermediary variables are only defined once and there is only one way to do so). 
* It is already the case : the system check with hasattribute if the system has an order. If it doesn't, it recreates one (but not necessary)
* Yes, I think getting the results as the dictionnary results solve that problem already

Didou09:
>     * The first point should not be true : all working order are giving the same system (because intermediary variables are only defined once and there is only one way to do so).

You're right, my mistake

>     * It is already the case : the system check with hasattribute if the system has an order. If it doesn't, it recreates one (but not necessary)
> 
>     * Yes, I think getting the results as the dictionnary results solve that problem already
Yes indeed

Didou09:
Done with:
* debugging `__repr__()`
* setting `dtime/dt = 1` instead of `dt`

Tests are passing.
=> once you are done with the remaining PEP8 issues in Main.py and Overlay.py, I guess you can finally merge :-)

DaluS:
It seens that there are two small issues : 
* **solver meed specification** If you don't specify 'GK' as a model the system cannot run. Which is surprising as if we load 'GK' it works...
* **The timestep message** is not working well, I typically have this when using my IDE (spyder). The run is also extremely slow, I assume due to this terminal operations. when verb=0 the system is as fast as the original, so this print operation is very slow (maybe we should show the number of iteration every second of the run or something like that ? This print is the most computer-demanding operation of the code right now :D )
image

The rest seems to work well ! Here are some ideas : 
* `_class` is the real core of the program, so I think its real name should be _CORE or _GEMMES_CORE 
* I think `_class._class_checks.models.get_available_models` is great, we should have more of this. Typically an overlay that tell yours for each model : 
    * Its description
    * The list of variables
    * the list of parameters
    * The equations with their comment
    * The different set of pre-loaded parameters one should try
* `sol.get summary` should give the initial value of intermediary functions
* In print `None` should be replaced by an empty space

DaluS:
I have added to the system :

** In `_def_fields.py` 
* More comments 
* Put the warning operation in `check_DFIELDS` a function
* Added a flag a the beginning `__DOTHECHECK`
* I have put all the variables which are not yet in a model in the unused file (we will put them back with the models)

DaluS:
The test are not working because there is no `main.py` but there is no `main.py` in this architecture... We should soon redefine all the test on the new structure

Didou09:
OK,
I can't work on it today, but I will work on it tomorrow.

> solver meed specification If you don't specify 'GK' as a model the system cannot run. Which is surprising as if we load 'GK' it works..

Will fix tomorrow


> The timestep message is not working well, I typically have this when using my IDE (spyder). The run is also extremely slow, I assume due to this terminal operations. when verb=0 the system is as fast as the original, so this print operation is very slow (maybe we should show the number of iteration every second of the run or something like that ? This print is the most computer-demanding operation of the code right now :D )

Will implement the possibility of printing only N % of the time steps tomorrow

Regarding the way it is printed in spyder, and the fact it is very slow with verb=1: it is due to Spyder, not to the code.
Indeed, IDE likes Spyder or Pycharm are very good for developing (that's what they are optimized for), but they are not so good for executing. Because they want to provide all those extra functionalities to the user, they mess up with the way the commands are executed, and quite often they also mess up with the plots. I don't know how exactly, but Spyder probably tries to intercept the prints at some point, making them slow and possibly modifying the final display a bit.
My advice: develop in an IDE (if you want), but execute in a simple python or ipython console, with no IDE.
This is also what users will do: users will either use a python/ipython console or even execute straight from bash without even opening a python console.

If you try to time the execution time of the run with verb = 0 and 1, on an ipython console, it's almost the same:
image


> ** Diagnostic :** GK_Reduced (which is a Goodwin reduced in fact :) , even simpler to verify its property !) Is not taking the values in _def_fields

Will fix tomorrow

> [Edit : I found why there was this value of 0.1, 0.1, which are initial values in the model. I am against, it is either in _def_fields, or be in a dictionnary of initial parameters/variable. Not hidden that way as it can be defined at multiple place by the developper which is I think a bad idea]. The quality of the Rk4 seems surprising (timestep dt = 0.01 ):

I am not sure I understand: the 0.1 was in the model file right ? in model/_model_GK_Reduced.py in the field 'initial' ? This is not hidden since it is the place where the user will define his/her model, so he has full control of it. Also, I might have forgot but I don't remember that we defined a dictionary of initial values.
In the current version, initial values are a field in the dictionary defining the model (_DPARAM).
A key defining an ode, has a field 'func' for the equation, a field 'initial', for the initial value, and all other fields defining the type of equation, the type of quantity, etc, so that all aspects of a quantity (equation + initial value in particular) are in the same place.
Do you mean you want to put the initial value in a separate place in a different dictionnary ? It's fine for me but where would you like to put it ? 
I can work on it tomorrow too, but please make sure to clarify your needs before so I don't go in the wrong direction.


> This should be an oscillation and not a divergence ( I have to check if the performance of the previous version was equivalent, I just remembered a better convergence on dt)

Hum, difficult to tell for me as I have not worked a lot with the previous version
=> That's where detailed unit tests would be useful to detect that kind of differences
=> This will be possible as soon as I am done with the data saving issue (we'll be able to save the result of a simulation we are sure of, and use it to check all following versions of the code reproduce the same result).
Can you take care of this comparison berween the convergence of the rpevious version and this one or shall we work on it together tomorrow afternoon ?


> _class is the real core of the program, so I think its real name should be _CORE or _GEMMES_CORE
> I think _class._class_checks.models.get_available_models is great, we should have more of this. Typically an overlay that tell yours for each model :
> Its description
> The list of variables
> the list of parameters
> The equations with their comment
> The different set of pre-loaded parameters one should try
> sol.get summary should give the initial value of intermediary functions
> In print None should be replaced by an empty space

Ok, will take care of all this tomorrow

> In the check of allowed fields, latex writing (typically ^{ }) should not be taken into account, and combinations of them should be allowed (Typically W/m2 should be written W.m^{-2} and valid with only W and m defined as acceptable units)
_DFIELDS should have autocompletion : if you don't fill a field for a variable it should be default things going ( None, empty strings etc)

I agree this would be nice, but I think this is lower-priority and should be adressed later in separate issues.

> I don't understand why you have not defined phi0 and phi1 value with a lambda equation. They are not variable and should not be in the model. They are fixed parameters, it is simply more convenient to define phinull than the two others

If you look in the model files 5which have priority over _def_fields.py and which are the files which the user will look), it's exactly what is goig on: phi0 and phi1 are defined as lambda functions from phinull.
Their existence in _def_fields.py is just a legacy of the time when I was not sure what you prefered to do (because at some point you had written phi0 and phi1 as paarameters i one of the models, so I was not sure what you wanted to do. 
=> We can absolutely set them to lambda functions in _def_fields.py if you want, please confirm that's what you prefer and I'll deal with it tomorrow

> I think the classification of groups should be discussed, It is complex to choose precisely the one we should define or not... Especially some variable will couple groups
> dimensions should work in latex formalism : Temperature rate is a temperature per unit of time... dimension and unit should be unified

Agreed, that's typically the kind of normative decisions you should make, you can open a dedicated issue for discussion, we'll discuss it, and at some point you'll have to take a normative decision that won't be perfect but that you think is the best compromise

> lambda functions should be accepted through _DALLOWED_FIELDS

Will look at this tomorrow

> Check_DFIELDS should give the back the key that is not working also

Will look at this tomorrow

> We can add in the model a list of presets that function that way ( I put this in _model_G_Reduced) :

Agreed, will try to implement a functioning version tomorrow


Question:
-----------

* The aboves point are a mixture of urgent and less urgent matters
* You should decide:
     - which are so urgent that they should be done inside this particular (already big) PR ?
     - which should be treated separately in a dedicated PR after this one has been merged ? (keep in mind it is always better to work with smaller PR)
Obviously we have to fix the tests before merging, but on my opinion, appart from that, almost all the rest can be treated separately

DaluS:
## Point by point discussion

* **About IDE problem** I understand that IDE cause problem in that type of case, it is just that I know most of the actual user of the code are working with IDE such as spyder or Pycharm (I had the same type of problem few years ago in atom with fancy terminal messages). 
* **About Main.py and overlay.py** They are not compatible with the current version of the code (actually the only file which are compatible are the one not in the folder `NotIncluded`. In consequence we should not care about them yet, they are depreciated and should be incorporated in the new architecture. Only files outside of this folder should be checked ! Same thing for the unitary test of main : it is not in the environment anymore. I replace `overlay` by the `cockpit` file which behave kinda the same :) but different names as, if there is a continuity in the philosophy, there is no continuity in its development
* **About value priorities in model and** _def_fields : A "naive" user (as I was yesterday) would assume that the model is the logical core (the ensemble of equations) and that the fields contains the value of the system (parameters and initial conditions). Actually the value of lambda can be defined in both of them, but the system takes the initial value contained in the `_model` file. That architecture mean that we cannot compare model on a same set of parameters, and in consequence it depreciate the value of `_def_fields` (which is a common space for all models). Moreover, the core as written actually cannot function without the initial conditions told in `_model` : I think the idea of overloading `_def_field` with infos in the model is great, but it is not what is happening here. It looks like the initial value of ODE variable are never read in `_def_field`. 
* **The solution about this double-place of definition** is that inside the model we add a dictionnary of dictionnary (as written in `G_reduced` and if the user call one of these dictionnary the values inside will be loaded on top of `_def_field`. It allows : 
    * Comparison between models (no dictionnary load)
    * Run of typical situations (as for example the one presented in the linked article for example)
* **about the dt convergence** I'll doublecheck the previous version in the day, totally agree that there should be unitary test later on for this 
* **about the case of Phi0 and Phi1 described as intermediary variables in the model** The trouble I have with this writing is that it looks like those are variable, which they are not. They are parameters deduced from parameters, and as such should not be recalculated nor stocked at every iteration. I think it is reasonable to have parameters that are deduced from other parameters. Physically, it's much more convenient only to define `PhiNull` as it is the variable that makes sense. 
> => We can absolutely set them to lambda functions in _def_fields.py if you want, please confirm that's what you prefer and I'll deal with it tomorrow
I confirm ! It's just that I tried and couldn't do it ( I might have done it wrong) 
* **About the fields in _def_fields** : Sure, I'll open an issue on it !

------------

## Finish this PR

The reason I have not merged this branch into devel is that :
* I think I should have merged what I had done (with overlay and everything in main) into the main branch as "This is the sum of all the functionality in the code" and stop the developement of it on devel. Then we decide that the new architecture is the one we should develop on. 

** Why do I think that ** 
* This branch has fundamentally a totally different structure as the previous code ( and again I think this structure is great ! ) 
* A lot of the previous functionality (plots, multiple systems in parrallel, existence of a main that allows modification of the system...) are not usable in this version. 
* If we merge that code we "erase" ( I know that with the Github architecture we don't but you see the spirit) the previous architecture and functionality. In consequence for me this code is a lot of step forward for a clean, strong basis BUT it is, for the moment, a step backward in term of functionality and practicability. I'm pretty sure this will soon no longer be the case, but these depreciation are pushing me to be careful on the subjet. 

**What we should do ?** 
I'll create a branch that correspond to the legacy version with full functionality based on the previous core, then merge this one into devel ! We should just keep in mind that those will need to be added asap so that the functionality of the code goes along with its evolution

Didou09:
> About IDE problem 

ok, but if you really want the prints to work fine with IDEs (despite the IDE slowing execution), we'll have to open a dedicated issue for this, and since I don't have an IDE you'll probably have to be in charge of it. A parallel solution would be to show and explain everyone that an IDE is not necessary.

> About Main.py and overlay.py

ok

> The solution about this double-place of definition

ok, do I understand correctly if I say that in order of priority for taking the value of a parameter or a function should be:
Preset dict > model > def_fields.py

> about the case of Phi0 and Phi1 

I will move them to _def_fields.py to avoid all ambiguity.
But just to make sure this is clear: currently (i.e.: in this PR), the code is already doing what you want: they are not treated as intermediary function but as parameters that depend on another parameter.
I put the eqtype '_Intermediary_' just because we had not discussed that particular case and I wasn't sure how you wanted to call that them.
But in fact, in a preparation phase before running, the code automatically detects that they only depend on fix-value parameters and replaces the function by the computed value. So they are not treated as variable and are not stored / recomputed at every iteration. They are really treated as fix-value parameters. It is just a way of telling the code to compute them from phinull.

> About value priorities in model and _def_fields

Ok, agreed.
But, again, to make it clear for this particular PR (and maybe explain something that might have been confusing, sorry):
* originally _def_fields.py was a common place for all models so that users don't have to redefine the dimensions, units, symbol... of all quantities they use. As its name suggests, it is only a place to hold defaults fields (in the case in which they are not defined in `_model_....py`), like a back-up that, on my opinion, the user should not have to edit.
* A quantity either has a fixed 'value' or has a 'func' (not both). If it has a 'value' it is a fixed parameter, if it has a it can be a time-dependent variable (ode with 'initial' value or intermediary or auxiliary function without initial value) or also a fixed parameter (if the function only depends or other fixed parameters.
* So for 'lambda', defined as a function in _model_....py, the 'func'+'initial' defined by the user takes over and the 'value' in _def_field.py is not read. This way the user only has one file to edit, not 2.
* I guess a simple measure to avoid all confusion is to remove the field 'value' from _def_fields.py ?
* Maybe a possible source of confusion is that you might have understood 'value' as a way of defining the 'initial' value. In the current version these are 2 different things: either 'value' or 'func'+'initial'.
* I implemented a default fixed 'value' in _def_fields.py for lambda because I did it for all quantities, since some of them can be both fixed-value parameters in some models and time-varying variables defined by 'func' in other models.

This is the way it is done in this particular PR (my apologies if I did not make clear the difference between 'initial' and 'value' in our last meeting). But of course, this is only a PR and all of this can be modified in later dedicated issues.

I think I have understood your points, I'll try to make things clearer in futur PR.

> This branch has fundamentally a totally different structure as the previous code ( and again I think this structure is great ! )
> A lot of the previous functionality (plots, multiple systems in parrallel, existence of a main that allows modification of the system...) are not usable in this version.

I think I understand what  you mean (i.e.: you cannot use these previous functionalities with the new structure).
However, if I may, I would like to emphasize one point that is important for collaborative coding: the previous functionalities are all still usable with the previous structure. It is not a detail, and I did my best to make sure of that and to not break anything, to be sure that you would still be able to do some research with the previous tools (if necessary) while at the same time being able to start testing the new structure.
I mention this because I took care of it and because it is important in terms of methodology: this is how things should be done when adding new features: not breaking anything. A new structure should not replace and old one (and break it...) it should coexist side by side until they are slowly merged, stable PR after stable PR. This way everyone can still work although the architecture is slowly changing. This is also what ensures that you can merge the PR without fear of loosing anything, and why I kept the unit tests for the Main.py.
In that sense I am not sure it is a good thing to remove the unit tests for Main.py. It's a choice that has its reasons, but by doing this it prevents us from keeping an eye on the old functionalities (they could be developped in parallel, and its always good to check something works before we slowly integrate it into a new structure). I'm ok if you still want to remove these tests, but keep in mind the methodology and try to decipher how you want us to work regarding these aspects for future PRs.  
Creating a branch for the previous version is also a possibility, but in my view it is non-necessary (since everything was still working). Branches are typically for adressing a specific issue, like a bug or a new functionality, here it will probably only be used as back-up.

Issue #48, opened 2021-07-15
IMPORTANT ISSUE : ALL MODIFICATION BEFORE V1.0


DaluS:
# Here we are ! 

There is a new core in this model ! This core is based on three files : 
* a core with most operations inside, in `_class.py` 
* a model (there are plenty of them) in the folder `models` with the moment ( each in a separate file `_model_NAMEOFTHEMODEL.py` )
* a list of fields in the folder `models` called `_def_fields` that contains all the keys necessary for the system to roll  

An additional file `cockpit.py` can be used to control the core and do what you want (because a pirate is free) with it. The file contains the explanation on how to make it work 

This new structure changes a lot for the best, and should be stable for now. A few fields or functions should move, but users/developer should not see too many of them. 

**Where are we heading ?** The goal is now to implement all the basic ideas we have and the previous functionality before the core change. When this will be done, it will be a new milestone and **GEMMES 1.0** will be published 🥳  

The road toward `V1.0` is still long. See the roadmap to get the journey

--------------------
What we should do right now : 
* Edit this list if something is missing 
* Convert into an issue a point when you want to solve it 
* If all the points are completed, get a bottle of champomy and merge the program on `Main` branch to celebrate `V1.0`
--------------------

# An ordered list of issues to be tackled (roadmap)

### In model files `_model_NAME.py`

- [x] functions should not have an 'initial' key, as it is in _PRESET
- [ ] Parameters deduced of parameters should be defined in the `_def_fields` by default, and their definition inside the model is only to overwrite it 
- [x] The list of parameters in a `_model_name` should be : 
     - [x] a list of all of them
     - [x] **better** : recreated by the system and not necessary defined
- [x] `_FUNC_ORDER` should not have to be defined (the system check if the attribute exist) 
- [x] `_DESCRIPTION` should be loaded in the core (if defined as _FUNC_ORDER, warning if not existing), and accessible to the user
- [x] `_PRESETS` should be a dictionnary of dictionnary, each subdictionnary can have two entries :  ( 'fields' with all the parameters/value changes , 'com' a commentary on what does the model. ). It should be loaded in the core, the list of preset accessible to the user, and overload `_def_fields` if the user ask to
- [x] The parameters that are deduced from parameters should be defined as such in the dictionnary (lambda value definition) not in the model (or have their definition overloaded if necessary) [edit :] We can add a new equation-type which is parameter ! 

### In `_def_files.py` 

See #47 for this mostly !
    
#### Content and elements

[...]

PR #50, opened 2021-07-15
Paul improve

DaluS:
Improvement of the code according to #48 

In it I added 
*


DaluS:
If we want to be pep8 compliant (which on many of the criteria is for me a reduction of readability, but can be discussed) we can do a run using typically https://pypi.org/project/autopep8/ if we really think that it is an issue ( I personally think that would only be necessary before `V1.0` release

DaluS:
@Didou09 I let you confirm the merge and check the changes, I'll be available after 5pm french time if you want to discuss ! please check #48 that will be (if you agree) the roadmap for the following weeks

models/_def_fields.py
```diff
@@ -56,21 +56,12 @@
...
     'value': None,
     'com': None,
-    'dimension': [
-        'time', 'time rate',
-        'Temperature', 'Temperature rate',
-        'Gas concentration', 'Gas flux',
-        'Power flux',
-        'Humans',
-        'Money', 'Real Units'
-    ],
```
Didou09, line 66: Question:  Are you sure you want to remove dimension? (just to be sure)

models/_def_fields.py
```diff
@@ -56,21 +56,12 @@
...
-        'y', 'y^{-1}',                          # Time
-        'Dollars',                              # Money
-        'C', 'C^{-1}', 'C^{-2}', 'C^{-zeta}',   # Concentration
-        'Humans',                               # Population
-        'W/m2',                                 # Linked to temperature
+        'Units',  # 
+        'y',      # Time
+        '$',      # Money
+        'C',      # Concentration
+        'Humans', # Population
```
Didou09, line 64: Question: units 'Units' ? 
Do I understand correctly that this is like a placeholder for yet-undecided units ?

Didou09 (approved):
I approve the PR,
I just have have 2 non-critical questions (see below)
They are for my information and do not block my approval

this a nice small and understandable PR, thanks

Didou09:
Comments for next time:
---------------------------------

* Try to choose a more explicit name for the PR ('Paul improve' is very non-explicit: the fact it is your work is already known on Github, and the fact it is an improvement probably applies to all PR...) If possible try to refer to an existing associated issue, by number

* Try to put more details in the description, in particular, try to include:
    - a section description of the reasons / motivation behind this PR
    - a brief summary of the main changes
    - a list of the issues adressed
    - if there a new functionality: a minimum working example
 
* Regarding PEP8 compliance: I merged because I understand you don't want to spend time on it and prefer to leave it to v 1.0. I don't really think this is a good idea (we'll have a lot of work at the end, and it keeps you from taking good habits), but you're the lead and I can live with that choice if that's what you want.

Didou09:
We usually delete the branch after the merge, but this is your branch so I don't delete it and let you choose if you want to delete / keep it

models/_def_fields.py
DaluS: Yes I'm sure !

models/_def_fields.py
DaluS: It will be "technological units" or "real units" later on, that is what is commonly used in litterature to count machine and not the money they are worth

Issue #53, opened 2021-07-16
Incorporate intermediary function first steo into reset()


Didou09:
Motivations:
----------------
* In PR #50, you introduced a method `Solver.runIntermediaryAtInitialState()` that computes the value of the intermediary functions for the first time step. I understand this can be useful for the user so when he/she print  a `Solver.get_summary()`, he/she will see the inital state not only for ode but for all time-varying quantities.
* I agree it is a good idea, and I propose to include this into the `Solver.reset()` method, as this method already sets the first time step of ode, it seems natural that it would set the first time step for everything.
* Also, it would minimize the number of different methods the user has to deal with.

Propose changes:
------------------------
* I will submit a PR incorporating the content of this method into `reset()`

DaluS:
Done !

PR #54, opened 2021-07-16
Issue053 intermediary initial step in reset

Didou09:
Motivations:
----------------
* Introduction in issue #53 
* The idea of computing the first time step of all functions introduced by @DaluS in PR #50 is extended here, the associated function is incorporated into an section of `Solver.reset()`, as I assumed it would be logical and perhaps simpler for the user.
* As a result of this (pre-computing t`=0` for intermediary functions), the solving order for time step ii in run() has to be reversed: before it was intermediary then ode, now it must be ode then intermediary
* I also propose several minor changes and code clean-up, some of which result in a minor acceleration of the simulation time :-)

Main changes:
--------------------
* A few redundant lines removed in `Solver.__init__()`
* `Solver.runIntermediaryAtInitialState()` is deleted and the code is incorporated in `Solver.reset()`
* In `Solver.run()`, ode are now solved before intermediary
* In `Solver.run()`, there is now an option (compute_auxiliary) to activate / deactivate on-the-run computation of auxiliary functions (set to `True` by default, but that can be changed)
* Since now a computation of the dict of input `kwdargs` for each function is necessary both in `reset()` and in `run()`, I propose to compute this dictionary in `self.set_dparam()`, once and for all, as a hidden attribute. This way is called in `reset()` and `run()` instead of being re-created in each of them (faster, simpler, single-sourced). Also it now contains directly the values to be used and not just the keys.
* This it also allows the `lambda` exception to be dealt with there instead of dealing with it at every time step inside the time loop, which seems to result in a _non-negligibly faster run time_ :-)

Issues:
---------
* Fixes issue #53 

Results / examples:
--------------------------
* unit tests unchanged and passing
* user -interface unchanged (no new method, but all you wanted to do is done)
```
In [1]: import _class

In [2]: sol = _class.Solver('GK')
['Y', 'L', 'Pi', 'lambda', 'omega', 'phillips', 'kappa', 'I']
```

At this point the first time step of all functions has been computed
```
In [7]: sol.get_summary()
Numerical param.  value  units  comment                                        
----------------  -----  -----  -----------------------------------------------
Tmax              100    y      Duration of simulation                         
dt                0.01   y      Time step (fixed timestep method)              
nt                10000  None   Number of temporal iteration                   
nx                1      None   Number of similar systems evolving in parrallel

Model param.  value                  units   group       comment                                                    
------------  ---------------------  ------  ----------  -----------------------------------------------------------
r             0.03                   y^{-1}  Prices      Interest at the bank                                       
k0            -0.0065                None    Keen        Percent of GDP invested when profit is zero                
k1            0.006737946999085467   None    Keen        Investment slope                                           
k2            20                     None    Keen        Investment power in kappa                                  
alpha         0.02                   y^{-1}  Population  Rate of productivity increase                              
delta         0.005                  y^{-1}  Capital     Rate of capital depletion                                  
beta          0.025                  y^{-1}  Population  Rate of population growth                                  
nu            3                      None    Production  Kapital to output ratio                                    
phinull       0.04                   None    Philips     Unemployment rate that stops salary increase (no inflation)
phi0          0.04006410256410257    None    Philips                                                                
phi1          6.410256410256412e-05  None    Philips                                                                

function  source                         initial    units               eqtype  comment                                           
--------  -----------------------------  ---------  ------------------  ------  --------------------------------------------------
a          alpha * itself                1.00e+00   Output Humans^{-1}  ode     Exogenous technical progress as an exponential    
N          beta * itself                 1.00e+00   Humans              ode     Exogenous population as an exponential            
K          I - itself * delta            2.70e+00   Real Units          ode     Capital evolution from investment and depreciation
W          itself * phillips             8.50e-01   Dollars             ode     Wage evolution through phillips curve             
D          I - Pi                        1.00e-01   Dollars             ode     Debt as Investment-Profit difference              
Y          K / nu                        9.00e-01   Real units          inter   GDP in output quantity                            
L          K / (a * nu)                  9.00e-01   Humans              inter   Workers                                           
Pi         Y - W * L - r * D             1.32e-01   Dollars             inter   Absolute profit                                   
lambda     L / N                         9.00e-01                       inter   employment rate                                   
omega      W * L / Y                     8.50e-01                       inter   Wage share of the economy                         
phillips   -phi0 + phi1 / (1 - lamb)**2  -3.37e-02  y^{-1}              inter   Wage inflation rate                               
kappa      k0 + k1 * np.exp(k2*Pi/Y)     1.20e-01                       inter   Part of GDP in investment                         
I          Y * kappa                     1.08e-01   Dollars             inter   Investment                                        
g          (1 - omega) / nu - delta      4.50e-02   y^{-1}              aux     Relative growth                                   
d          D / Y                         1.11e-01                       aux     relative private debt                             
pi         1 - omega - r * d             1.47e-01                       aux     relative profit                                   
i          Y * 0                         0.00e+00   y^{-1}              aux     Inflation rate                                    
time       1.                            0.00e+00   y                   ode     Time vector          
```    

And are all re-computed using `sol.reset()`

The execution time of the simulation, which was, on my PC, around **1.30 seconds** in the previous version (see https://github.com/DaluS/GEMMES/pull/46#issuecomment-880465885), is now faster both with and without computing auxiliary functions on-the-run, with respectively a **20 %** and **28 %** gain:

```
In [5]: %timeit sol.run(compute_auxiliary=False)
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
931 ms ± 9.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit sol.run(compute_auxiliary=True)
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
time step 10000 / 10000
1.03 s ± 6.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/_class.py
+++ b/_class.py
@@ -18,10 +18,8 @@ class Solver():
     def __init__(self, model=None):
         self.__dparam = {}
         self.__func_order = None
-        #if model is not None: ### COMMENTED FOR NO DEFAULT MODEL ACCEPTANCE
         self.set_dparam(dparam=model)
-        self.__run = False
-        self.runIntermediaryAtInitialState()
+
     # ##############################
     #  Setting / getting parameters
     # ##############################
@@ -46,9 +44,9 @@ class Solver():
 
         """
 
-        # If all None => set to self._PARAMSET
+        # If all None => set to self._MODEL
         c0 = dparam is None and key is None and value is None
-        
+
         if c0 is True:
             print()
             dparam = self._MODEL
@@ -98,7 +96,37 @@ class Solver():
             dparam=dparam, func_order=func_order, method=method,
         )
 
-        # reset variable
+        # store a simplified dict of variable arguments 
+        # used in reset() and run()
+        lode = self.get_dparam(eqtype='ode', returnas=list)
+        linter = self.__func_order
+        laux = self.get_dparam(eqtype='auxiliary', returnas=list)
+
+        self.__dargs = {
+            k0: {
+                k1: self.__dparam[k1]['value']
+                for k1 in (
+                    self.__dparam[k0]['args']['ode']
+                    + self.__dparam[k0]['args']['intermediary']
+                    + self.__dparam[k0]['args']['auxiliary']
+                )
+                if k1 != 'lambda'
+            }
+            for k0 in lode + linter + laux
+        }
+
+        # Handle the lambda exception here to avoid test at every time step
+        # if lambda exists and is a function
+        c0 = (
+            'lambda' in self.__dparam.keys()
+            and self.__dparam['lambda'].get('func') is not None
+        )
+        # then handle the exception
+        for k0, v0 in self.__dargs.items():
+            if c0 and 'lambda' in self.__dparam[k0]['kargs']:
+                self.__dargs[k0]['lamb'] = self.__dparam['lambda']['value']
+
+        # reset all variables
         self.reset()
 
     def get_dparam(self, verb=None, returnas=None, **kwdargs):
@@ -153,7 +181,7 @@ class Solver():
     def reset(self):
         """ Re-initializes all variables
 
-        Only the first time step (initial values) of ode is preserved
+        Only the first time step (initial values) is preserved
         All other time steps are set to nan
         """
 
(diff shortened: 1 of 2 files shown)


Didou09:
Conflicts solved @DaluS , you can merge if you agree
Have a nice weekend

DaluS:
Perfect Thank you ! Have a nice weekend :)

PR #56, opened 2021-07-16
Pipify

DaluS:
* Added a Autopep8.py that autopep the content of a folder 
* Renamed Solver to Hub 
* Clean of the set_dparam ( check of input data in `_class_check` with function `_checkInputDparam`, more concise writing at a few places)

Brain washed out by this week, I'll be better next one !


NotIncluded/Autopep8.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+Created on Fri Jul 16 14:59:42 2021
+
+@author: Paul Valcke
+"""
+
+import autopep8
+import os 
+
+files = [ f for f in os.listdir() if '.py' in f]
```
Didou09, line 11: Deatils:
' avoid one-letter variables (`ff` instead of `f`)
* Safer and more modern syntax `.py in ff`  => `ff.endswith('.py')`

models/_def_fields.py
```diff
@@ -34,46 +34,46 @@
 
 import numpy as np
 
-__DOTHECHECK = True
+__DOTHECHECK = False
```
Didou09, line 37: Are you sure you want to deactivate a safety check by default before we have fixed all inconsistencies?

Didou09 (approved):
Fine for me, I approve the PR

I have some comments / questions (see below), but they are not critical

NotIncluded/Autopep8.py
DaluS: > Deatils:
> ' avoid one-letter variables (`ff` instead of `f`)
> 
> * Safer and more modern syntax `.py in ff`  => `ff.endswith('.py')`

Ok for both ! I thought :
* In list comprehension or onelineloop that's ok (Will do ff or equivalent next time) 
* I didn't knew endswith ! I'll use that in the future

models/_def_fields.py
DaluS: For the moment it doesn't bring any information as we are not sure about what field/structure we want. For example, a run right now gives a lot of empty field, or issue related to the "latex formalism" 
image

I think we can reactivate it once it brings the real information !

NotIncluded/Autopep8.py
Didou09: There is also str.startswith() that can be useful :-)

models/_def_fields.py
Didou09: ok :+1:

Didou09:
Closed ?

So you did not want to merge it ?

Issue #59, opened 2021-07-21
Rewriting the task-list


DaluS:
# Formalism

## Formalism for `_def_fields` 

- [x] `_DFIELDS` is a dic(A) of dic(B) of dic(C), with each dic(C) being a field.
- [x] Each field need a (`value` or "func", or 'ode' and 'value'), a `units`. It can have a comments, a symbol 
- [x] The dimensions can be deduced from the units 
- [x] The group is deduced from dic(B)
- [x] two different dic(B) cannot contains the same value 

- [x] Intensive/extensive/dimensionless should be deduced from the system

## Formalism and properties for the `_model` file 

- [x]  Must be able to define new fields as in `_def_fields` (at the cost of a warning)
- [x]  Must work as definition overload of `_def_field` 
- [x]  The `_DPARAM` is in two dictionnaries (One for the ODE,  One for the state functions, PDE and DDE would have their own) , with a function and a comment each time. 
- [x]  The list of parameters should not be necessary (deduced from the model) but can be given 
- [x]  The order of intermediary functions should not be necessary but can be given 
- [x]  There should be a loadable description
- [x] There should be a dictionnary of Preset with the following form : 
![image](https://user-images.githubusercontent.com/11523050/126520537-14d9148f-5b76-4afd-a099-fdb0a00d53c5.png)
- [x] The preset should be loadable from a hub

DaluS:
Reintegrated to the path of `V1.0`

PR #61, opened 2021-07-21
New formalism for `_Def_Fields` and `_model` file

DaluS:
## First the miscellaneous modification 

1. Changed again (a merge might have not functionned properly) the "Hub" name of the core
2. Modified a part of the _PRINT_OR_WAIT_ to function correctly in the case of float number ( some others are still not working) 
3. Added a basic function to describe each model and the presets in __init__ of models called `_core._class_checks.models.describe_available_models()`
4. 
image

## Modification on `_def_fields_NEWFORMALISM` 

Correspond to modification of #59 , the file itself is not added in the workflow but is ready to. 
The main properties : 
* `_DLIBRARY` is the one accessible for the modeler (easy to modify/implement new fields
* The group classification is caused by the organization of the library ( dict of dict ), allow folding in IDE also. 
* Dimensions are now a dictionnary with in keys : 
    * multiplier (to handle Gigaton, 10^12 etc after)
    * each unit (y, dollars etc) are keys and the exponent is the value ( example productivity )
image
* Ode, state function, parameters have a different writing 
    * parameters 
image

    * ode  
image

    * functions 
image

* There will be functions added to transform a parameter into a state function (easy), a parameter into an ODE (easy), a state function into an ODE (more complex)

## MODIFICATIONS ON `_model_GK-NEWFORMALISM`

* _DPARAM will be transformed into _LOGICS, as a dictionnary of dictionnaries, with one section for `ode`, one for state functions `func`, other fields for DDE and PDE will be added later when needed
* `_FUNC_ORDER` and `_FUNC_PARAMS` will not need to be filled

## WHAT TO COME  

* Transforming CORE so that he is able to read both files
* Adding the overloading elements to the class 
* Adding overload-parrallel system (transformation to N systems in parrallel) 
* Enjoying it as it will be fully functional with a consistent formalism that allows future development :)


Didou09:
> Changed again (a merge might have not functionned properly) the "Hub" name of the core

As mentionned in the PR #56  itself: you did not merge, you closed the PR, without merging, that's why

models/_model_GK-NEWFORMALISM.py
```diff
@@ -0,0 +1,150 @@
...
+
+import numpy as np
+
+
+# ---------------------------
+
+
+_FUNC_ORDER = None # user-defined function order (optional)
+
+_FUNC_PARAMS = None # user-defined list of parameters (optional)
```
Didou09, line 16: But you don't specify the list of parameters here for that model, where is it specified?

The user might not want to specify all parameters in the preset, only those the values of which he wants to change

There should be somewhere a table listing the parameters for each model, if it is not here, where do you want to put it ?
We could try to decipher it dynamically from the functions if you don't want to have it stored, that's a possibility I guess

models/_def_fields_NEWFORMALISM.py
```diff
@@ -0,0 +1,389 @@
+# -*- coding: utf-8 -*-
```
Didou09, line 1: What's the difference between this file (_def_fields_NEWFORMALISM.py) and the previous one (_def_fields2.py) ? Which one is the new one ?

Didou09 (commented):
I guess this is work in progress since the tests are not passing.
So I am not approving the PR at this point, just providing an intermediate review with comments / questions.

I see you point in the model files, you want to separate the set of equations from the parameters and initial values. I agree this is a good move.

However, I don't see (yet) where you want to define the list of parameters for a given model?

Also, I would probably remove remove the field 'value' from _def_fields.py, in order to separate what is likely to be changed by the user ('value') from is likely not to be changed ('units', com'), so the user does not try to edit _def_fields.py but only the model file (the presets).

This would allow to define the defaults values in model (where the user can change them) at the same time as defining the list of parameters.

What do you think ? Maybe a visio today afternoon would help the discussion ?

DaluS:
> > Changed again (a merge might have not functionned properly) the "Hub" name of the core
> 
> As mentionned in the PR #56 itself: you did not merge, you closed the PR, without merging, that's why

Oh I didn't understood that ! 
It's same issue for the unit test not working : they try to test the class solver and not hub (I'll fix that) 

> However, I don't see (yet) where you want to define the list of parameters for a given model?
> But you don't specify the list of parameters here for that model, where is it specified?
>  There should be somewhere a table listing the parameters for each model, if it is not here, where do you want to put it ?
We could try to decipher it dynamically from the functions if you don't want to have it stored, that's a possibility I guess

It should not be necessary to precise parameters. When checking the model (all the lambda functions and all the keys in the dictionnary), we can determine the functions (called in lambda, in keys) and parameters (called in lambda, not in keys). 
That might be dangerous I agree, so maybe using the previous parameters definition (then check consistency with above elements) might be safer (but more tedious for modeler). a `_PARAMETERS`.

image

> The user might not want to specify all parameters in the preset, only those the values of which he wants to change

As a preset overload `_dparam`, we only have to state the points we want to modify ! That's exactly the idea and the formalism for the moment in 

> What's the difference between this file (_def_fields_NEWFORMALISM.py) and the previous one (_def_fields2.py) ? Which one is the new one ?

I will remove `_def_fields2` it's depreciated :) I forgot about it ! 


> Also, I would probably remove remove the field 'value' from _def_fields.py, in order to separate what is likely to be changed by the user ('value') from is likely not to be changed ('units', com'), so the user does not try to edit _def_fields.py but only the model file (the presets).
> This would allow to define the defaults values in model (where the user can change them) at the same time as defining the list of parameters.

I think the library should have default value ( putting ode is not a good idea for example) so that a user that has no Idea of one value could use this default one.

Didou09:
@Dalus, ok for a meeting tomorrow,

Ho by the way, I don't want to emphasize only the points that deserve discussion, this new feature (`_core._class_checks.models.describe_available_models()`) you introduce is a very good idea I think, congratulations ! (sorry, I should have mentionned it earlier)

Didou09:
> Oh I didn't understood that !

Remember you can always re-open the [PR](https://github.com/DaluS/GEMMES/pull/56) and merge it, but first you need to restore the branch (because apparently you deleted it).

If you want we can do this together tomorrow

DaluS:
This is suprising, it looks like I broke `_model_GK`, (the rest is working), I don't know what happened :) 

I'll check that tomorrow ! No need to merge (it's a draft)

--------------------------------

What's in : 

* An update to the readme
* A few updates on the `_NEWFORMALISM` that is ready to be inserted into the system
* An ensemble of functions I'd like to implement in the core 

What's next : 
* Repair `_model_GK`
* finish the PR 
* rewrite the necessary part of the `_core` to implement the newformalism
* add the functionalities that are in `cockpit` in the comment section 

```
for NameOfTheModel in ListOfModels :
    sol = _core.Hub(NameOfTheModel)
    PRESETS = []
    for preset in PRESETS :
        sol = _core.Hub(NameOfTheModel)
        
        #sol.load_preset(preset)
        
        sol.load_field({'a' : 1})
        sol.load_field({'a' : [1,2,'lin']})
        sol.load_field({'a' : [1,2,'log'],
                        'alpha' : 0.03})
        sol.load_field({'nx': 100})   
        sol.load_field({'a': [1,2,3]}) 
        
        sol.get_summary()
        sol.run(solver='eRK4-homemade', verb=1.1)
        
        sol.getCycleAnalysis(key='lambda')
        sol.getCycleAnalysis(key=False)
        
        Result = sol.get_dparam(returnas=dict)
                
        plt.figure()
        plt.subplot(211);plt.plot(Result['time']['value'],Result['omega']['value']);plt.ylabel('Omega')
        plt.subplot(212);plt.plot(Result['time']['value'],Result['lambda']['value']);plt.ylabel('Lambda')
        plt.suptitle(NameOfTheModel+' '+preset)
        plt.show()
        
        sol.plot('all')
        
```

Didou09:
Pro tip: to solve conflicts (because devel has changed a bit), go on your local devel branch, update it (git pull origin devel), then go on your local branch NewFormalism and merge devel into it (git merge devel), so you'll solve conflicts locally, then push to the remote NewFormalism branch

DaluS:
The reason why it is not working is because the previous initial condition was very bad - W=0.1 implied that 13% of the initial production went to salary in that case, which caused investment to be huge and send the system in a weird place. 
In consequence, the system is saying that the result differ which is true : the initial conditions in the program (with a depreciated writing in the future) are different, but better to test the system :)

Issue #3, opened 2021-06-28
Output data saving


Didou09:
Motivation:
--------------
* It seems that the code currently allows to save the modules themselves (function Miscfunc.savedata())
* In particular, it saves Main.py and FunctionsGoodwin.py, presumably to save the input parameters stored in these.
* As far as I understand it, the output data does not seem to be saved 
* In principle, storing a copy of the code is already done by git and github

Proposition:
-----------------
* In terms of general strategy, and architecture, for traceability, reproducibility and ease of use, I would recommend to:
   - Make sure all input parameters can be put into a single input dict
   - Make sure the solver / main function can take that input dict 
   - Make sure all produced results can be put into a single output dict
   - implement the possibility of saving that output dict, into a npz file
   - Implement an explicit naming convention for saved files: typically:
         `Input_ParmeterName1ParameterValue1_ParameterName2ParameterValue2_..._ParameterNameNParameterValueN.npz`
        `Output_ParmeterName1ParameterValue1_ParameterName2ParameterValue2_..._ParameterNameNParameterValueN.npz`
      Where only the most relevant parameters are included (file names are often limited in length)

DaluS:
Yes this saving method is weird and in fact I never used it. I think I'll add a method in the class to : 
* Stock all the necessary data generated by a run
* Save the class object with the new values
* Expand back the data when needed

About parameters and input, for me there are three different type of values (hence 3 dictionnary for the moment) : 
* the "physical/behavioral" parameters (price markup, efficiency of capital...)
* the numerical parameters ( Number of system solved in parrallel, time increment...)
* the initial conditions of the system
I like the three of them to be different, but we can regroup them into one if you think that's more suitable. Or having a dictionnary of dictionnary which regroup all of them ? 

For saved files I think it'd be complex to find the right naming convention. We can ask the user to name the run and put this as the name, with a default name otherwise (typically the date) and a .txt of the parameters on the side (that can be easily read)

DaluS:
As mentionned in #48 , we should decide on : 
* The type of data we want to save
* Its format
* The structure of saving folders 

Before coding it !

DaluS:
Done

PR #62, opened 2021-07-22
Issue003 output data saving

Didou09:
Motivation:
-------------
* This is a first version of a data saving / loading procedure, operational and tested
* It is exhaustive (saves everything)
* It is tested in unit tests (they have been updated) for all available models
* Unit tests also check that the saved data and the loaded data are identical (no loss or change of information)

Main changes:
----------------
* A new `utilities/_saveload.py` module is created, where one will find all routines associated to saving / loading
* A new method `Solver.save()` has been implemented
* Several options are available:
    -` Solver.save(path=...)` allows to choose in which folder to save (by default => `output/`)
    - `Solver.save(name=...)` allows the user to define a custom name for the file
      The custom name is inserted inside a name template which is as follows (as you requested @Dalus if I remember correctly):
             `Output_MODELNAME_name_USER_DATE.npz`
    - By default the method prints a line saying where exactly the file was saved.
    - If not specified, `name` is by default `'placeholder'`
    - `Solver.save(fmt=...)` allows the user to decide under which format to save the file.
      By default only `'npz'` is available for now (we'll see later if there is a need for others)
* A new routine `_core._saveload.load()` now allows to load a previously saved file
* New unit tests have been implemented to test saving and loading
* In addition, to check that the loaded data was identical to the saved data in unit tests, a the method `Solver.__eq__()` was overloaded to allow for in-depth comparison of all attributes to tell whether 2 instances are identical or not
    - If they are identical (same attributes)  `obj1 == obj2` returns `True`
    - If they have different attributes `obj1 == obj2` returns `False `and also lists all the attributes that are different
    - This allows very easy and fast verifications
    - To do it t was necessary to: 
         - group all attributes into dictionaries (dparam, which already existed, and dargs and dmisc, names can be changed later if necessary)
         - Implement methods for in-depth comparisons between dict (in `_utilities._class_utility.py`)
         - Implement hidden methods for:
              - exporting the instance as a dict: `Solver._to_dict()`
              - creating an instance from a dict: `Solver._from_dict()`
* These new methods make life easier and it allowed for the easy creation of a new replication method: `Solver.copy()` returns an instance identical to the original one, but still as a distinct object that can be modified without affecting the original.
* A new unit test was added to test that new `Solver.copy()` method

All unit tests (old and new) are passing :-)

Issues:
-------
* Fixes Issue #3 

Typical use case:
------------------

```
In [1]: import _core                                                                                                                                

In [2]: sol = _core.Solver('G_Reduced')                                                                                                             
['pi', 'phillips', 'g']

In [3]: sol.run()                                                                                                                                   
time step 10000 / 10000

In [4]: sol.save()                                                                                                                                  
Saved in:
	/Home/DV226270/Projects/GEMMES/output/Output_G-Reduced_placeholder_DV226270_20210722-150807.npz

In [5]: sol_load = _core._saveload.load('output/Output_G-Reduced_placeholder_DV226270_20210722-150807.npz')                                         

In [6]: sol == sol_load                                                                                                                             
Out[6]: True

```

Note that the saved file, with everything included, is less than `1 Mo` (~`864 Ko`), so at this point we can definitely afford to save everything.

More detailed example:
--------------------------

Here we modify a parameter of `sol_load` to check the behaviour of the `Solver.__eq__()` method

```
In [10]: sol_load.set_dparam(key='alpha', value=0.01)                                                                                               

In [11]: sol == sol_load                                                                                                                            
The following differences have been found:
	- dargs['lambda']['g']: different values
	- dargs['omega']['phillips']: different values
	- dargs['pi']['omega']: different values
	- dargs['phillips']['lamb']: different values
	- dargs['g']['pi']: different values
	- dmisc['run']: different values (True vs False)
	- dparam['alpha']['value']: different values (0.02 vs 0.01)
	- dparam['lambda']['value']: different values
	- dparam['lambda']['source_kargs']: different values (itself=0, g=0, alpha=0.02, beta=0.025 vs itself=0, g=0, alpha=0.01, beta=0.025)
	- dparam['omega']['value']: different values
	- dparam['phillips']['value']: different values
	- dparam['g']['value']: different values
	- dparam['pi']['value']: different values
	- dparam['time']['value']: different values
Out[27]: False
```

We notice that the list of differences include the changed value of `alpha` but also:
- all value arrays (because they were all reset)
- The default values of lambda functions that take `alpha` as a parameter
- `run` is now `False` (due to the reset)

Let's now copy one of our instance and check it is identical to the original while still being a distinct object

```
In [10]: sol2 = sol.copy()                                                                                                                          

In [12]: sol2 == sol                                                                                                                                
Out[12]: True

In [13]: sol2 is sol                                                                                                                                
Out[13]: False
```
This way it becomes very easy to dynamically replicate an instance, modify the parameters of the replica and re-run it with modified parameters, as illustrated below:

```
In [14]: sol = _core.Solver('G_Reduced')                                                                                                            
['pi', 'phillips', 'g']

In [15]: sol.run()                                                                                                                                  
time step 10000 / 10000

In [16]: sol2 = sol.copy()                                                                                                                          

In [17]: sol2.set_dparam(key='alpha', value=0.015)                                                                                                  

In [18]: sol2.run()                                                                                                                                 
time step 10000 / 10000
```

We now have 2 instances that ran with 2 different sets of parameters, for the same model.

### DIFF
--- a/_core.py
+++ b/_core.py
@@ -322,7 +322,7 @@ class Solver():
         ar2 = [
             tuple([
                 k0,
-                v0['source'].split(':')[-1].replace('\n', '').replace(',', ''),
+                v0['source_exp'],
                 "{:.2e}".format(v0.get('value')[0, idx]),
                 str(v0['units']),
                 v0['eqtype'].replace('intermediary', 'inter').replace(
--- a/_core.py
+++ b/_core.py
@@ -6,7 +6,7 @@ import numpy as np
 import time
 
 # Library-specific
-from utilities import _utils, _class_checks, _class_utility
+from utilities import _utils, _class_checks, _class_utility, _saveload
 
 
 class Solver():
@@ -18,6 +18,8 @@ class Solver():
     def __init__(self, model=None):
         self.__dparam = {}
         self.__func_order = None
+        self.__model = None
+        self.__solver = None
         self.set_dparam(dparam=model)
 
     # ##############################
@@ -173,6 +175,11 @@ class Solver():
     def func_order(self):
         """ The ordered list of intermediary function names """
         return self.__func_order
+
+    @property
+    def model(self):
+        """ The model identifier """
+        return self.__model
 
     # ##############################
     # reset
@@ -339,6 +346,7 @@ class Solver():
     # ##############################
     # run simulation
     # ##############################
+
     def run(self, compute_auxiliary=None, verb=None):
         """ Run the simulation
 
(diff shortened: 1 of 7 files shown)


Didou09:
@Dalus, please note that the remaining PEP8 syntax error is irrelevant:

DaluS:
Fantastic ! I really like it. 
I don't understand exactly what is `placeholder` ? Perfect for the rest :D

Didou09:
> Fantastic ! I really like it.
> I don't understand exactly what is `placeholder` ? Perfect for the rest :D

Glad you like it :-)

placeholder is just a default name in the case the user does not provide one, hence it acts like a place-holder (it holds the place of the name).
You can change that default name if you want (or I can do it in my next PR, just let me know)

Issue #60, opened 2021-07-21
_run_check not working for verb=0


DaluS:
when run with 'eRK4-homemade' and `verb=0`, the system crash

![image](https://user-images.githubusercontent.com/11523050/126568276-6ac7f54e-635d-4937-9f6e-a450c8faa522.png)

problem solved with another value of verb

PR #63, opened 2021-07-23
Issue060 verb0

Didou09:
Motivation:
-------------
* It was reported in issue #60  that the `Solver.run()` method was failing with `verb=0`
* This PR fixes it and also implement **_updated unit tests_** to make sure it is tested and does not happen again

Main changes:
----------------
* The issue came from the fact that `end`, `flush` and `timewait` were not defined in `utilities/_class_checks._run_check()` when `verb = 0` (because they were defined inside the `if` statement that start from `verb = 1`).
* They are now defined ahead of the `if` statement (set to `None` because they are not used when `verb = 0`)
* When such a bug is identified, it means the unit tests were not complete enough because they did not detect it earlier
=> It is then good practice to update the unit tests to make sure they would detect it if it happened again
=> The unit tests are updated to test all possible values of `verb`

Issues:
-------
* This fixes issue #60 

Example:
----------

```
In [1]: import _core                                                                                                                                

In [2]: sol = _core.Solver('G_Reduced')                                                                                                             
['pi', 'phillips', 'g']

In [3]: sol.run(verb=0)       

In [4]:
```

### DIFF
--- a/tests/test_01_Solver.py
+++ b/tests/test_01_Solver.py
@@ -107,9 +107,17 @@ class Test01_Run():
         lsolvers = ['eRK4-homemade', 'eRK2-scipy', 'eRK4-scipy', 'eRK8-scipy']
 
         # list of entry parameters to try
-        for model in self.dmodel.keys():
-            for solver in lsolvers:
-                self.dmodel[model].run(solver=solver)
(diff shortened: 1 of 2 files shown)


DaluS:
Perfect thanks !

PR #72, opened 2021-07-28
Analysis of Cycles

DaluS:
# Adding the analysis of Cycles

This model does include economic cycles, which are "fast" oscillations, on top of a slower evolution. This PR allow the study of both phenomena at the same time. 
It also includes the plot that show it for each variables 

The methodology used is the local detection of maximum on one curve (homebrewed) with a wraparound

**IMPORTANT : THIS MODULE DOES NOT WORK WITH MULTIPLES SYSTEMS IN PARRALLEL FOR HE MOMENT**

image

## The modifications 

* A few plots have been improved, especially `plots.Var(hub, 'lambda', idx=0, cycles=True, log=False)`
    * log put the figure with y axis as logarithmic
    * cycles add the cycle analysis on the figure
    * idx is the index of the system
* Three functions have been added to the core : 
   * `FillCyclesForAll(self, ref=None)`, The big wrapup to apply the logic on all variables
   * `FillCycles(self, var, ref='lambda')` calculate the variables once the cycles are detected. 'ref' 
   * `findCycles(self, refval)`  calculate the cycles

## How to access these terms : 

run `hub. `
now in `__dparam[key]['cycles']` you get keys : 
* 'period_indexes' list of list of the borders of each cycles
* 'period_T_intervals', same but with time
* 't_mean_cycle', time in the middle of each cycle
* 'period_T', duration of each cycle
* 'reference', the name of the variable cycles have been calculated on
* 'meanval', mean value during each cycle
* 'medval', median value during each cycle
* 'stdval', standard deviation value during each cycle
* 'minval', min value during each cycle
* 'maxval' max value during each cycle


_core.py
```diff
@@ -447,14 +444,134 @@ def run(
...
+        '''
+        Detect all positions of  local maximums
+        '''
+        # initialisation
+        Periods = []
+        id1 = 1
+
+        self.__dparam[refval]['cycles'] = {}
+
+        dic1 = self.__dparam[refval]['cycles']
```
Didou09, line 543: `dic1` here seems to be defined but not used (and re-defined below)

_core.py
```diff
@@ -470,7 +587,7 @@ def _to_dict(self):
         }
         return dout
 
-    @classmethod
+    @ classmethod
```
Didou09, line 590: There has been a long debate about decorator syntax in python, as far as I know the reference syntax so far does not include a space between the `@` and the decorator name. Can you explain why you inserted one?

_core.py
```diff
@@ -447,14 +444,134 @@ def run(
...
+'minval': [],   # minimal value in the interval
+'maxval': [],   # maximal value in the interval
+'reference': ref,  # the variable that has been used to detect cycle
+}
+            '''
+
+        # check if reference has already calculated its period
+        # the reference has cycle and this cycle has been calculated on itself
+        dic1 = dic['cycles']
+        Ready = False
```
Didou09, line 492: By convention, no upper case for variables: `ready`

_plots.py
```diff
@@ -63,23 +63,68 @@ def plotVar(sol, key, idx=0):
 
     '''
     plt.figure('key :'+key+'| system idx'+str(idx), figsize=(10, 5))
+    fig = plt.gcf()
```
Didou09, line 66: Directly store the figure at instanciation:
`fig = plt.figure('key :'+key+'| system idx'+str(idx), figsize=(10, 5))`

Didou09 (commented):
In principle, the PR is good

I have a few non-critical comments regarding the code (see below), you can decide to take them into account and update the code, or not (they are non-critical).

However, the good practice is that every time you introduce a new feature (method, function), you also implement the corresponding unit test to make sure it works in all common cases.

You implemented 5 new methods / functions, but no unit tests:
* `Hub.FindCycles()`
* `Hub.FillCycles()`
* `Hub.FillCyclesForAll()`

* `_plot.Var()`
* `_plot.phasespace()`

You should at least implement:
* 3 unit tests for the 3 methods (or a unique unit tests testing all 3 since they are similar / coupled)
* One unit test for each new plotting method

Alternatively, if you don't have time to do the unit tests now, you can say it in the PR and immediately open a new issue dedicated to these unit tests to keep track of them and make sure we don't forget them

Also it is good practice to include, in the PR description, one exmaple plot for each new plotting function introduced (`phasespace` is missing).

_core.py
```diff
@@ -447,14 +444,134 @@ def run(
...
+        #
+
+        # Fill for each the characteristics
+
+    def findCycles(self, refval):
+        '''
+        Detect all positions of  local maximums
+        '''
+        # initialisation
+        Periods = []
```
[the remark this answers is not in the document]
Didou09: Done

_core.py
```diff
@@ -447,14 +444,134 @@ def run(
...
+
+        dic1 = self.__dparam[refval]['cycles']
+        val = self.__dparam[refval]['value']
+
+        # identification loop
+        while id1 < len(val)-2:
+            if (val[id1] > val[id1-1] and
+                    val[id1] > val[id1+1]):
+                Periods.append(1*id1)
+            id1 += 1
```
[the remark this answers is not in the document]
Didou09: Might propose a modification in a later dedicated PR

_core.py
Didou09: Might address in a later dedicated PR

_core.py
Didou09: Done

_core.py
Didou09: Done

_plots.py
Didou09: Done

Didou09:
Since you are on holidays, I took the liberty of fixing the easiest points above (syntax: minor case instead of upper case...) directly in new commits into this PR, so I can merge it is merged before you come back from holidays

RThese are very minor changes (click on the last 5 commits to see the details) that do not affect what you have done
I also fixed the PEP8 non-compliances.

PR #73, opened 2021-07-28
Better get summary

DaluS:
Now, if the system did run, there is a column "final" added : 

When you execute : 
`hub.get_summary()
hub.run(verb=1.1)
hub.get_summary()`

You get : 
image

Nothing more needed :)


Didou09 (commented):
The change regarding the get_summary() is ok

However, you created a branch based on you previous PR branch (`OscillationAnalysis`).
Better practice would have been to create this branch directly from `devel`, because:

* The changes in `OscillationAnalysis` and in this branch (`BetterGetSummary`) are independent
* So the corresponding PR should also be independent
* In the case the reviewer has comments on the previous PR (`OscillationAnalysis`), but agrees with this PR, he cannot accept it directly because he would implicitely also accept the previous PR (`OscillationAnalysis`) on which he has comments

By branching from successive branches instead of branching from `devel` you make all your PR dependent on each other:
They cannot be accepted independently.

For merging this one, which is fine for me as a reviewer, we will have to wait until the previous one is merged (on which  have comments).
It is not a big deal this time because my comments are non-critical, but if they had been critical, this PR would have been blocked too. And removing critical comments (e.g.: debugging) can sometimes take days or weeks, so it could potentially block this PR for days or weeks, even though it is perfectly fine.

Didou09:
Since you are on holidays, I take the liberty to implement myself some (very minor) changes to the syntax to make it a bit more compact.
I also added 2 more info in the `get_summary()`: the run variable and (if solver has been run), the solver name:

image

DaluS:
Cool ! Thanks :) 
Also do you have an idea on how we can have a practical convention in terms of scientific/decimal notations ? I have no idea what we should do, anyway this is good enough for the moment :)

Didou09:
> Cool ! Thanks :)
> Also do you have an idea on how we can have a practical convention in terms of scientific/decimal notations ? I have no idea what we should do, anyway this is good enough for the moment :)

No simple idea now, I guess the current status is satisfactory and changing it is probably a low-level priority compared to other issues.
I open a dedicated issue to address this later if you agree?

PR #74, opened 2021-07-28
All phase space plot

DaluS:
Added in `_plots.py` a new one. 

It functions with 

`dimensionlessnumbers = ['omega', 'lambda', 'd']
plots.AllPhaseSpace(hub, dimensionlessnumbers, idx=0)`

`dimensionlessnumbers` contains here all variables which are indeed dimensionless (here done manually). 

The system does all combination of these variables and then look at their properties : 
image


Didou09 (commented):
Fine for me,

With the same comment as for the previous PR: better practice would be to branch from devel

Or, if this PR depends on a previous PR, branch from it and only from it (the PR  BetterGetSummary) is included here, but it is completely independent and should not be included here. It is due to the fact you branched successively from the same branch:

`devel` => `OscillationAnalysis` => `BetterGetSummary` => `AllPhaseSpacePlot`

You should have done either:
`devel` => `OscillationAnalysis` 
`devel` => `BetterGetSummary` 
`devel` => `AllPhaseSpacePlot`

Or, if necessary (if this PR depends on `OscillationAnalysis`, but I' m not sure it is the case):
`devel` => `OscillationAnalysis` => `AllPhaseSpacePlot`
`devel` => `BetterGetSummary`

DaluS:
> Fine for me,
> 
> With the same comment as for the previous PR: better practice would be to branch from devel
> 
> Or, if this PR depends on a previous PR, branch from it and only from it (the PR BetterGetSummary) is included here, but it is completely independent and should not be included here. It is due to the fact you branched successively from the same branch:
> 
> `devel` => `OscillationAnalysis` => `BetterGetSummary` => `AllPhaseSpacePlot`
> 
> You should have done either:
> `devel` => `OscillationAnalysis`
> `devel` => `BetterGetSummary`
> `devel` => `AllPhaseSpacePlot`
> 
> Or, if necessary (if this PR depends on `OscillationAnalysis`, but I' m not sure it is the case):
> `devel` => `OscillationAnalysis` => `AllPhaseSpacePlot`
> `devel` => `BetterGetSummary`

The second case would be better here, however I find it in fact a confusing way. Especially because this has been mostly tangled development, with a change of the core and rewriting multiple functions for the new formalism... Honestly I prefer to base in core changes the development based on the previous development, it's makes more sense in my opinion.

Didou09:
Merging and opening issue #82 on corresponding unit tests simultaneously

PR #75, opened 2021-07-31
Change fields formalism

DaluS:
Not finished : 
* It happens that there is a problem with set new params individually, according to the unitary test. 
* Regression test do not work, which is normal


DaluS:
Reset function do no reset effectively the system (It cannot rerun)

This should be checked before Merging ( I might have forgotten some fields update in the reset)

DaluS:
Depreciated (Integrated in another PR)

PR #76, opened 2021-08-04
New core set dparam

DaluS:
Will be written later. (need to leave work)
I know unitary test wont work, they should be updated

Basically totally changed set_dparam so that it is developper friendly in its structure, many task are now automatic, and handle new formalism well


DaluS:
Depreciated (Integrated in another PR)

PR #79, opened 2021-08-05
Future main

DaluS:
This is the latest improvement of the code. 

There are 3 important things : 
* Unitary test are outdated, will not work here and that is normal
* The RK4 is not working 
* The parrallelisation is not working yet (but quite close to, most is written)

To see the characteristics, the best is to check the `Tutorial.ipynb` which contains everything


DaluS:
Depreciated ( added in another PR)

PR #90, opened 2021-08-17
Issue89 code independent files

Didou09:
Motivations:
----------------
* Trying to separated PR #75 , #76 and #79  into smaller independent parts
* This PR focuses on adding the files from those PR that do not directly affect the code
* I did not really review them since this is mostly documentation (that can evolve) and I focus on the lower-level tasks
* Since they cannot affect the code or unit tests directly, it is safe to merge this PR

Proposition:
----------------
* It is best practice to not put `ipynb` into the code directly but rather in a directory dedicated to documentation
* Thus I created a `/doc` folder where I put all `ipynb` and where I suggest we put all documentation-related files
* Later we can set-up automated documentation using [sphinx](https://www.sphinx-doc.org/en/master/) or [doxygen](https://www.doxygen.nl/index.html), these specialized libraries typically work with a dedicated `/doc` directory.


DaluS:
Yes ! That looks like good practice indeed
Good, I'll check sphinx and doxygen once the core is well written also :)

Issue #87, opened 2021-08-12
Reviewing PR #75 #76 #79


Didou09:
Motivation:
---------------
* PR #75, #76, #79 introduce new formalism and several other changes
* They are big (30 files changed)
* I need a bit of time to review it all seriously, to do so I will:
        - copy the branch on my local
        - Go through your changes one by one, to understand exactly what is done
        - If necessary, propose small changes to make sure it passes unit tests (so you don't have to do it)
        - If necessary update the unit tests
* If I have small changes to suggest I will propose them through a PR that will be a duplicate of #75, with a  few differences (so I don't affect your PR, which is preserved as such, and you can choose which one you want to merge)

Starting today

DaluS:
Super cool ! 
Do not hesitate to contact me anytime if something is unclear. I was thinking about restructuring a bit this part to make future add easier ( especially the conditions on "ode" "statevar" or the autofill which exist at multiple places ), I'm eager to discuss about it if you like !

Didou09:
Today I only had ~2 hours this morning, so I won't be available in ~5 min (my girlfriend just woke up, breakfast time together :-))

But maybe we can arrange a short visio tomorrow morning ? somewhere between 08h30-11h ? Since I am on holidays I try to work only while she sleeps.

First quick question already:
--------------------------------------
From reading the new model files, I understand there is no `auxiliary` function anymore, they are included as `statevar` together with the former `intermediary` functions right ?
If you don't have time to answer we can discuss it tomorrow (or later)

I estimate I will need a few days to finish the review of those PR. 
Maybe a full week depending on my availability, is that timeline fine for you ?

P.S.: I read the presentation you gave on 08.08.2021 (I downloaded the slides / ipynb), I'm ok with everything you presented :-)

DaluS:
I hadn't seen this message, sorry ! 
I "removed" auxilliary because it's far from priority (it will be useful when spatial PDE or a way more variables than what we have right now), I tried not to remove the code but simply comment/not update it. It'll be great to develop it again as we add PDE into the code

DaluS:
Here it is ! That quite long issue is now fixed, thanks again @Didou09 !

Issue #58, opened 2021-07-21
Differences between solvers


Didou09:
Motivation:
--------------
* PR #57 showed a typical case in which the numerical behaviour depends on the solver chosen
* We need to understand these differences and, if possible, correct them (at leats for the homemade solver)

Example:
-------------
```
In [1]: import _core

In [2]: solH = _core.Solver('G_Reduced')
Suggested order for intermediary functions (func_order):
['pi', 'phillips', 'g']

In [3]: sol2 = _core.Solver('G_Reduced')
Suggested order for intermediary functions (func_order):
['pi', 'phillips', 'g']

In [4]: sol4 = _core.Solver('G_Reduced')
Suggested order for intermediary functions (func_order):
['pi', 'phillips', 'g']

In [5]: sol8 = _core.Solver('G_Reduced')
Suggested order for intermediary functions (func_order):
['pi', 'phillips', 'g']

In [6]: solH.run()
time step 10000 / 10000

In [7]: sol2.run(solver='eRK2-scipy')
The solver successfully reached the end of the integration interval.
Success: True
Nb. fev: 3004 for 10000 time steps (0.3004 per time step)

In [8]: sol4.run(solver='eRK4-scipy')
The solver successfully reached the end of the integration interval.
Success: True
Nb. fev: 6007 for 10000 time steps (0.6007 per time step)

In [9]: sol8.run(solver='eRK8-scipy')
The solver successfully reached the end of the integration interval.
Success: True
Nb. fev: 15016 for 10000 time steps (1.5016 per time step)

In [11]: k0 = 'lambda'

In [12]: plt.figure(); plt.plot(solH.get_dparam(returnas=dict)['time']['value'], solH.get_dparam(returnas=dict)[k0]['value'], '-k', label='homemade'); plt.plot(sol2.get_dparam(returnas=dict)['time']['value'
    ...: ], sol2.get_dparam(returnas=dict)[k0]['value'], '.r', label='scipy RK2'); plt.plot(sol4.get_dparam(returnas=dict)['time']['value'], sol4.get_dparam(returnas=dict)[k0]['value'], '.b', label='scipy R
    ...: K4'); plt.plot(sol8.get_dparam(returnas=dict)['time']['value'], sol8.get_dparam(returnas=dict)[k0]['value'], '.g', label='scipy RK8'); plt.legend(); plt.gca().axhline(0.97, c='k', ls='--'); plt.gca
    ...: ().axhline(0.94675, c='k', ls='--');
[...]

Issue #66, opened 2021-07-23
Start packaging


Didou09:
Motivation:
-------------
* We now have a code that has a few solid features
* Some of the routines, although very helpful (`get_available_model()`, `get_available_output()`, and others...) are not easy to find for the user, they have to dig into the hidden modules `_core._saveload`....
* I think the project is getting reasonably mature and, if you agree, we can start trying to package it like a real python library, and maybe even distribute it on [Pypi](https://pypi.org/) so users can install it directly from the web by doing `pip install gemmes` for example
* This way we can design the commands we want to accessible so a user can do:

```
import gemmes as gem

sol  = gem.Solver('GK')

gem.get_available_models()

gem.get_available_output()

...
```

If you agree to go in that direction, it will require a few changes in the architecture (nothing too impressive but still a few things).

I can take care of this issue, and try to do it progressively (through several small PR so you can follow how to do it).

 I wait for your validation before working on it.

DaluS:
@Didou09 This should wait, as for the moment we would prefer to keep it in the lab ( and publish the code with an article)

PR #91, opened 2021-08-17
Issue87 pr75

Didou09:
Motivations:
----------------
* In the process of reviewing PR #75, #76 and #79, I tried to isolate the main changes to the code, by topic
* The most important change desired by @DaluS is the shift to the new formalism (i.e.: a different formalism for the user to specify a model as a dict of sub-dict + a dict of presets)
* This PR focuses on this part: updating to new formalism, without loss of functionality and without breaking the tests
* A good fraction of the proposed changes are inspired from the PRs #75, #76 and #79, but some parts are also a bit different (to provide all functionnalities while preserving the low-level safety checks)

Main changes:
--------------------
* The new formalism is implemented
* `get_available_models()` and `get_available_output()` are more complete and harmonized
* The unit tests are all operational and test all presets of all models for all solvers. No unit test has been lost and several new ones have been implemented

Disclaimer:
---------------
* The huge number of files changed is ;mostly due to the fact we are now testing all presets of all models for all solvers, so there are many new reference files for non-regression tests. These are binary files that do not need reviewing.

Detailed changes:
------------------------
* `_def_fields.py`: 
    - now contains a `_LIBRARY` dict defining all quantities known to the code
    - From this `_LIBRARY`, a `_DFIELDS` dict is automatically derived, to avoid breaking backward-compatibility
    - Safety checks on both are activated by default
    - The old `_DFIELD` dict is still there for back-up but deprecated and unsused (good practice to keep it for a few PR and then delete it)
 * `model` files:
     - now contain a `_LOGIC` dict that contains the equations, classified by type (`ode`, `statevar`...) into sub-dict
     - also conatin  a `_PRESET` dict containing pre-set values for parameters
     - The user does not need to specify the list of static parameters anymore (automatically identified by the code, from the keyword args of the functions, and from the `_LIBRARY`)
     - The old `_DPARAM` is still there for back-up (good practice), but deprecated and unused.
* `_core.Hub`:
    - the `Hub` class now allows to:
        - directly load a model and a preset (`hub = _core.Hub(model='GK', preset='default')`)
        - or to load a different preset (change preset) after the model has a been loaded (`Hub.load_preset('default')`)
    - The definition of the `dargs` dict has been moved to `utilities._class_checks.py` to make the code on the Hub class more readable and lighter.
    - `Hub.dmodel` is a new dict that contains:
        - the model name, file, description, presets, chosen preset...)
    - All other (tested) methods have been updated accordingly (`get_summary()`, `__repr__()`, ...)
* `_class_check.py`:
     - Most of the safety checks previously implemeted are preserved, because more safety is better than less
     - But since it was mentioned they were not easy to understand, some documentation and more comments were added
     - The functionalities behind routines like `FindParameters()` were implemented, just with a slightly different name / structure
* `get_available_models()`:
   - A redundancy was observed between 3 routines (`get_available_models()`, `describe_ALL_available_models()` and `describe_available_model()`)
   - A unique routine is left (`get_available_models()`) that offers all the functionalities of the 3 previous routines
   - `get_available_models(details=True)` will give a full description of desired models
   - `get_available_models(returnas=dict)` will return a dict of models 
* unit tests:
    - They now test all presets of all models with all solvers
    - A new unit test file has been added to test `get_available_models()` and `get_available_output()`
    - All tests are passing

What has not been implemented (voluntarily):
-------------------------------------------------------------
* Since the Hub class is a major tool for user-interface, it is good practice to minimize the number of methods that it offers, in particular, it should not make visible to the user low-level methods, hence several methods present in PR #79 where not implemented and instead where hidden in `_class_checks.py`. If we really want to make them visible, it is easy to create a small method that calls the routine.
* All variable and method names were written using minor case, as is [good practice](https://www.python.org/dev/peps/pep-0008/#id45)
* `_get_DicOfSameUnits()` seems to implement a functionality that already exists in `Hub.get_dparam()` (to be double-checked later)
* changes to the solver (`_eRK4_paul()`) will treated in a separate PR in relation with issue #58 

What is not implemented yet (to do):
------------------------------------------------
The PR #75, #76 and #76 also include code changes regarding:
    - cycles analysis
    - plots of cycles and phase spaces
    - changes to Cockpit.py
These will be treated in later separate PRs 

Issues:
----------
* Partially fixes issue #87 

Example:
------------

import and get available models, with various options (with and then without all details)

```
In [1]: import _core

In [2]: _core._class_checks.models.get_available_models()

#################################################
################### DESCRIPTION OF G_Reduced

file: /home/didier/Documents/Projects/GEMMES-1/models/_model_G_Reduced.py
presets:
	- smallcycle    : gives simple stable sinusoidal oscillations
	- bigcycle      : give extremely violent stable oscillations
	- badnegociation: displace the Solow Point and allow big cycles with few harmonics
nb. of functions:
	- ode: 2
	- statevar: 3

------ description ----

ABSTRACT: This is the simplest economic post-keynesian model
TYPICAL BEHAVIOR : metastable oscillations around a solow point
LINKTOARTICLE : Goodwin, Richard, 1967. ‘A growth cycle’, in:
    Carl Feinstein, editor, Socialism, capitalism
    and economic growth. Cambridge, UK: Cambridge University Press.

Created on Wed Jul 21 15:11:15 2021
@author: Paul Valcke


#################################################
################### DESCRIPTION OF GK

file: /home/didier/Documents/Projects/GEMMES-1/models/_model_GK.py
presets:
	- default: This is a run that should give simple convergent oscillations
	- crisis : This is a run that should create a debt crisis
nb. of functions:
	- ode: 6
	- statevar: 12

------ description ----

DESCRIPTION : This is a Goodwin model based on extensive variables.
Inflation not integrated to the process
TYPICAL BEHAVIOR : Convergence toward solow point ( good equilibrium) or debt crisis
LINKTOARTICLE:

Created on Wed Jul 21 15:11:15 2021
@author: Paul Valcke

In [3]: _core._class_checks.models.get_available_models(details=False)
The following predefined models are currently available:
	- G_Reduced: ['smallcycle', 'bigcycle', 'badnegociation']
	- GK: ['default', 'crisis']
```

Now returned as a dict or a list, for one or all models

```
In [4]: dmodel = _core._class_checks.models.get_available_models(returnas=dict, model='G_Reduced')

In [5]: lmodel = _core._class_checks.models.get_available_models(returnas=list)
```

Similarily, the user can check which output files already exist in the reference directory for non-regression tests,
and they can also be retruned as a dict

```
In [6]: _core._saveload.get_available_output(path='tests/output_ref/')
File selection criteria:
	- path: tests/output_ref/

Available files matching criteria:
	- tests/output_ref/Output_GK_eRK4-scipy_122_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK4-homemade_110_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK2-scipy_121_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK4-homemade_120_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK8-scipy_113_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-homemade_0_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-scipy_22_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK2-scipy_11_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK2-scipy_101_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-scipy_2_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK2-scipy_111_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK2-scipy_21_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK2-scipy_1_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK8-scipy_3_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK4-scipy_112_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK2-scipy_31_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-homemade_30_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-homemade_10_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-scipy_32_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK4-homemade_100_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK8-scipy_23_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK8-scipy_123_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK8-scipy_13_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-scipy_12_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK4-homemade_20_didier_20210817-184207.npz
	- tests/output_ref/Output_G-Reduced_eRK8-scipy_33_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK4-scipy_102_didier_20210817-184207.npz
	- tests/output_ref/Output_GK_eRK8-scipy_103_didier_20210817-184207.npz

In [7]: dout = _core._saveload.get_available_output(path='tests/output_ref/', returnas=dict)
```

Now he can define a hub from a model (and optionally from a preset), he can change preset and all usual methods still work fine

```
In [8]: hub = _core.Hub('GK', preset='default')
['inflation', 'Y', 'L', 'GDP', 'Pi', 'omega', 'kappa', 'd', 'I', 'pi']

In [9]: hub.run()

In [10]: hub.load_preset('crisis')

In [11]: hub.run()

In [12]: hub
Out[22]: 
model  source                                                        nb. model param  nb. functions  run 
-----  ------------------------------------------------------------  ---------------  -------------  ----
GK     /home/didier/Documents/Projects/GEMMES-1/models/_model_GK.py  16               17             True

In [13]: hub.get_summary()
Numerical param.  value          units      comment    
----------------  -------------  ---------  -----------
Tmax              100            y          No comment 
dt                0.01           y          No comment 
nt                10000          undefined  Constant dt
nx                1              y          No comment 
run               True                                 
solver            eRK4-homemade                        

Model param.  value                  units      group               comment                               
------------  ---------------------  ---------  ------------------  --------------------------------------
lambda        0.97                   undefined  CORE                No comment                            
phillips      0.031160968660968562   y^{-1}     Salary Negociation  Wage increase rate through employement
alpha         0.02                   y^{-1}     CORE                No comment                            
beta          0.025                  y^{-1}     CORE                No comment                            
delta         0.005                  y^{-1}     CORE                No comment                            
nu            3                      undefined  CORE                No comment                            
r             0.0                    y^{-1}     Debt                No comment                            
phi0          0.04006410256410257    undefined  Salary Negociation                                        
phi1          6.410256410256413e-05  undefined  Salary Negociation                                        
k0            -0.0065                undefined  Investment          No comment                            
k1            0.006737946999085467   undefined  Investment          No comment                            
k2            20                     undefined  Investment          No comment                            

function   source                      initial    final      units                     eqtype    comment                                           
---------  --------------------------  ---------  ---------  ------------------------  --------  --------------------------------------------------
a          alpha * itself,             1.00e+00   7.54e+00   Units.Humans^{-1}.y^{-1}  ode       Exogenous technical progress as an exponential    
N          beta * itself,              1.00e+00   1.26e+01   Humans                    ode       Exogenous population as an exponential            
K          I - itself * delta,         2.70e+00   1.32e+00   Units                     ode       Capital evolution from investment and depreciation
D          I - Pi,                     1.00e+01   2.75e+02   $                         ode       Debt as Investment-Profit difference              
W          itself*phillips,            8.50e-01   2.01e+01   $                         ode       salary through negociation                        
p          itself*inflation,           1.00e+00   1.00e+00   $                         ode       NO INFLATION FOR THE MOMENT                       
Y          K / nu,                     9.00e-01   4.40e-01   Units.y^{-1}              statevar  Leontiev optimized production function            
GDP        Y*p,                        9.00e-01   4.40e-01   $                         statevar  Output with selling price                         
inflation  0,                          0.00e+00   0.00e+00   y^{-1}                    statevar  INFLATION NOT CODED                               
L          K / (a * nu),               9.00e-01   5.84e-02   Humans                    statevar  Full instant employement based on capital         
Pi         Y - W * L - r * D,          -1.65e-01  -8.98e+00  $                         statevar  Profit for production-Salary-debt func            
omega      W * L / Y,                  8.50e-01   2.67e+00   undefined                 statevar  wage share                                        
kappa      k0 + k1 * np.exp(k2*Pi/Y),  -6.33e-03  -6.50e-03  undefined                 statevar  Relative GDP investment through relative profit   
I          GDP * kappa,                -5.69e-03  -2.86e-03  $                         statevar  Investment value                                  
d          D / GDP,                    1.11e+01   6.24e+02   undefined                 statevar  private debt ratio                                
pi         1 - omega - r * d,          -1.83e-01  -2.04e+01  undefined                 statevar  relative profit                                   
time       1.,                         0.00e+00   1.00e+02   y                         ode       dt/dt=1, time as ODE     
```


DaluS:
Excellent! 
I would not activate non-regression tests for the moment as the solvers are still not working.

A few details : 
- On `_model_G_Reduced`, phi0 and phinull are not the same thing, the preset should be written again. The reason I had to do that is that defining a parameter which is a function of parameters (and not a state variable) is not working for the moment... The solution would be either : 
    - Considering that phi0 and phi1 are state **variables** (they were function of phi0)
    - Change the way loading is done
- The __repr__ of the hub can be more useful : 
    - I do not think the file location is useful ( since it's totally loaded into the hub anyway)
    - I think its important to separate ODE, statevar and parameters ( which is not done here, I implemented previously a routine to determine the parameters) 
    - It could be interesting to add the date of the first load to know which core version has been used ( with a flag in the core file)
- You removed the difference between `definition` and `com` which is important : `definition` is the definition of the field, `com` tells you what is behind the logic hypothesis. Both should exist as they have different purpose 

I think we should not update scipy solvers in test for the moment, they have an output formalism which is not working well, and they cannot have the parallel input (if you have not repaired it) 

The priority, again, is still the RK4 that should really be repaired ASAP ! 

I'll do a more detailed check of the implementation in the code

DaluS:
# Review for _def_fields.py

1. What the modeler should see first is the fields, not the default or the depreciated code which 
2. Most of the constraints here have should not be constraint or not like this : 
    * Why adding default value on keys that are deduced from the system ( like units, group, type... ) ? 
    * The units that are not in DEFAULTFIELDS should be accepted (maybe with a warning). Some user will have to call units specific to some models with no need to correct here 
    * The units can have any quantities in the power ( `y^{1.5}` should be valid )

The file like this feel unclean and I do not understand where are many of the properties that had been implemented ( for dimensions etc )

Didou09:
> I would not activate non-regression tests for the moment as the solvers are still not working.

Don't worry, non-regression tests can be updated as we fix the solvers, and theyt are also useful for other things: they help discover hidden bugs with the saving / loading routines, with the new / old formalism.... They have already been useful for that a few times.

>     * When `describe_ALL_available_models()` it'd be better to have preset, nb of functions and description with the same importance. The order is here in reverse ( first description, then nb of function, then presets)

OK, will do in this PR (converted to draft in the meantime)

> On _model_G_Reduced, phi0 and phinull are not the same thing, the preset should be written again. The reason I had to do that is that defining a parameter which is a function of parameters (and not a state variable) is not working for the moment... 

I have to disagree, it is working, just introduce `phi0` and `phi1` as `statevar` functions of `phinull` and the code will automatically compute them and convert them to parameters

I had to change `phinull` to `phi0` because `phinull` was not a known parameter in the `ode` and `statevar` functions, therefore the safety checks detected that the preset was setting a parameter not known to the model (as they should). This was a temporary fix.
I can do a more relevant fix by defining `phi0` and `phi` as functions of `phinull` and letting `phinull` in the the preset.

> The repr of the hub can be more useful :

    I do not think the file location is useful ( since it's totally loaded into the hub anyway)
    I think its important to separate ODE, statevar and parameters ( which is not done here, I implemented previously a routine to determine the parameters)
    It could be interesting to add the date of the first load to know which core version has been used ( with a flag in the core file)


Agreed, but this can probably be done on a separate dedicated PR ?
This one focuses on the new formalism, I suggest we try not to implement everything at once.
Regarding the version number, a clean way of doing it is first to do a proper packaging of the code (cf. issue #66)

> You removed the difference between definition and com which is important : definition is the definition of the field, com tells you what is behind the logic hypothesis. Both should exist as they have different purpose

OK, will do in this PR


> I think we should not update scipy solvers in test for the moment, they have an output formalism which is not working well, and they cannot have the parallel input (if you have not repaired it)

I suggest we keep them in the tests, and address the parallelism issue in a separate PR (in preparation) that will also update the tests if necessary. (Again, tests are useful also for other things that the ones they were primarily created for)

> The priority, again, is still the RK4 that should really be repaired ASAP !

I'm on it today, on a dedicated PR to be submitted soon

> What the modeler should see first is the fields, not the default or the depreciated code which

OK, will move deprecated variables to the bottom of the files in this PR


>   Why adding default value on keys that are deduced from the system ( like units, group, type... ) ?
Please be more specific (maybe add a capture of the code part you're referring to), I m not sure I understand exactly what you mean.

>     The units that are not in DEFAULTFIELDS should be accepted (maybe with a warning). Some user will have to call units specific to some models with no need to correct here
OK, will implement a warning in this PR

>     The units can have any quantities in the power ( y^{1.5} should be valid )
OK, will do in this PR

>     The symbol is autofilled with the name of the field (and it's working well with the plot, that's great ! )
That's already done and working in this PR, was this just a comment ?

> The file like this feel unclean and I do not understand where are many of the properties that had been implemented ( for dimensions etc )
I'm not sure exactly what you are referring to, maybe we can discuss it live next week ?

DaluS:
> > ```
> > * When `describe_ALL_available_models()` it'd be better to have preset, nb of functions and description with the same importance. The order is here in reverse ( first description, then nb of function, then presets)
> > ```
> 
> OK, will do in this PR (converted to draft in the meantime)
> But to be sure: you want first preset, then nb. of functions and then description ? This is what is already implemented
> Or do you want the opposite ?
> 
> Is this fine: description, then presets, nb. of functions and finally location ?
> image

Yes that's great ! I had the opposite hence my comment (might have tried an outdated version) 

> > On _model_G_Reduced, phi0 and phinull are not the same thing, the preset should be written again. The reason I had to do that is that defining a parameter which is a function of parameters (and not a state variable) is not working for the moment...
> 
> I have to disagree, it is working, just introduce `phi0` and `phi1` as `statevar` functions of `phinull` and the code will automatically compute them and convert them to parameters

Ok great ! It was not the case in the branch I did, if it can stay like this (with correct priority of files, not checked yet) that's perfect. 

> 
> I had to change `phinull` to `phi0` because `phinull` was not a known parameter in the `ode` and `statevar` functions, therefore the safety checks detected that the preset was setting a parameter not known to the model (as they should). This was a temporary fix.
> I can do a more relevant fix by defining `phi0` and `phi` as functions of `phinull` and letting `phinull` in the the preset.

Yes, that we will improve later. Not sure if this will be compatible with my writing of the core (if yours works well, then that's fine !)

> > The repr of the hub can be more useful :
> 
> ```
> I do not think the file location is useful ( since it's totally loaded into the hub anyway)
> I think its important to separate ODE, statevar and parameters ( which is not done here, I implemented previously a routine to determine the parameters)
> It could be interesting to add the date of the first load to know which core version has been used ( with a flag in the core file)
> ```
> 
> Agreed, but this can probably be done on a separate dedicated PR ?
> This one focuses on the new formalism, I suggest we try not to implement everything at once.
> Regarding the version number, a clean way of doing it is first to do a proper packaging of the code (cf. issue #66)

Yes I agree ! But if you check what I had done in `newbranch` I have already kinda done that, it is just a few lines to be added :) 

> > You removed the difference between definition and com which is important : definition is the definition of the field, com tells you what is behind the logic hypothesis. Both should exist as they have different purpose
> 
> I'm not sure what you are refering to, in `_def_fields.py`, both `definition` and `com` co-exist, none has been removed.
> I copy-pasted the `_LIBRARY` you propose in PR #79 , which only contains `definitions`, but then the safety checks add `com` to all fields (with a default value at 'No comment' taken from `__DEFAULTFIELDS`). You can still set the com for each quantity of relevant, but was it removed from `_LIBRARY` then ?

Ok, then it is simply `get_summary` that run with the old version ! I added an update on `newbranch` to have both

> > I think we should not update scipy solvers in test for the moment, they have an output formalism which is not working well, and they cannot have the parallel input (if you have not repaired it)
> 
> I suggest we keep them in the tests, and address the parallelism issue in a separate PR (in preparation) that will also update the tests if necessary. (Again, tests are useful also for other things that the ones they were primarily created for)

Alright ! 

> > The priority, again, is still the RK4 that should really be repaired ASAP !
> 
> I'm on it today, on a dedicated PR to be submitted soon

Perfect ! That'll be great to have this one we can rely on (which works well also in parallel system) 

> > What the modeler should see first is the fields, not the default or the depreciated code which
> 
> OK, will move deprecated variables to the bottom of the files in this PR
> > Why adding default value on keys that are deduced from the system ( like units, group, type... ) ?
> > Please be more specific (maybe add a capture of the code part you're referring to), I m not sure I understand exactly what you mean.

`__DEFAULTFIELDS = {
    'value': {
        'default': None,
        'type': (int, float, np.int_, np.float_, np.ndarray, list),
        'allowed': None,
    },
    'definition': {
        'default': '',
        'type': str,
        'allowed': None,
    },
    'com': {
        'default': 'No comment',
        'type': str,
        'allowed': None,
    },
    'dimension': {
        'default': 'undefined',
        'type': str,
        'allowed': None,
    },
    'units': {
        'default': 'undefined',
        'type': str,
        'allowed': [
            'Units',  #
            'y',      # Time
            '$',      # Money
            'C',      # Concentration
            'Humans',  # Population
        ],
    },
    'type': {
        'default': 'undefined',
        'type': str,
        'allowed': [
            'intensive',
            'extensive',
            'dimensionless',
        ],
    },
    'symbol': {
        'default': '',
        'type': str,
        'allowed': None,
    },
    'group': {
        'default': '',
        'type': str,
        'allowed': None,
        # 'Numerical',
        # 'Population',
        # 'Prices', 'Capital', 'Philips', 'Gemmes',
        # 'Keen', 'Dividends', 'Economy', 'Production',
        # 'Coupling',
        # 'RelaxBuffer',
        # 'Climate', 'Damage',
    },
}
`


> 
> > ```
> > The units that are not in DEFAULTFIELDS should be accepted (maybe with a warning). Some user will have to call units specific to some models with no need to correct here
> > ```
> 
> OK, will implement a warning in this PR
> 
> > ```
> > The units can have any quantities in the power ( y^{1.5} should be valid )
> > ```
> 
> OK, will do in this PR
> 
> > ```
> > The symbol is autofilled with the name of the field (and it's working well with the plot, that's great ! )
> > ```
> 
> That's already done and working in this PR, was this just a comment ?

It's a comment ! I really liked seeing that :) For next PR we should think about having a nice units plots ! 

> > The file like this feel unclean and I do not understand where are many of the properties that had been implemented ( for dimensions etc )
> > I'm not sure exactly what you are referring to, maybe we can discuss it live next week ?

in `def_fields` of the branch `NewMain` there is the following functions : 

* `from_Library_to_DFIELDS` (that does take this more minimal library and complete it)
* `detectdimension` That does transform the string of units into a dimension dict
* `detectype` which is mostly useless right now, and can be removed (the type field is just a consequence of the dimension and is not used for the moment...)

At term, the dimension field can be removed (units dict is much more practical) and type might be removable again. 

And with pleasure to discuss it next week !

DaluS:
It seems that initialisation is not calculating first values for the moment 
image

And cycles detection is not working for the moment either

Didou09:
At this point this is still a draft (work in progress to take into account your comments + a few other things)
Will probably finish tomorrow
cheers

DaluS:
I saw in your push that you have a "param" explicitly defined by the user in def_fields, are you sure it is a good practice here ? The system automatically recognize what are parameters from the model file in `new_main`, I don't see cases it is necessary to declare it beforehand ? (or when a parameter is a function of parameters ?)

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+MODELS:
+-------
+* Replaced _DESCRIPTION by __doc__
+* hub.model dict vs str ?
+* __ini__.PrintDFIELDS() and subsequent thnigs (_get_summary()) => To be done later
+* 'i' vs 'inflation'?
+* 'p' vs 'prices'
+* 'K' vs 'Capital'
+* 'D' vs 'Debt'
+* 'N' vs 'Population'
```
DaluS, line 16: All of these are the common notation in the literature, writing it differently would makes the translation harder, I think we should keep these names that will talk for all economist (that is the role of `definition` also: bring a more practical information).

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* ...
+* 'phinull' defined in G_REduced preset 'badnegociation', but not used...
+
+
+Class utility:
+--------------
+* _get_DicOfSameUnits() seems redundant with hub.get_dparam(unit=...)
+* fromOldToNewModelFormalism() seems incomplete and unused
```
DaluS, line 24: I think you're right, it is an artifact of previous version. All has been transferred into load model

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* 'phinull' defined in G_REduced preset 'badnegociation', but not used...
+
+
+Class utility:
+--------------
+* _get_DicOfSameUnits() seems redundant with hub.get_dparam(unit=...)
+* fromOldToNewModelFormalism() seems incomplete and unused 
+* 'func': funcs['logic']: 2 different names for the same thing!
+* Preserving _check_dparam because:
+    - more complete checks (better more than less)
```
DaluS, line 27: Better more than less is not necessary a good practice : I think better explicit than more I would say ? 
Although I understand that in this case we should keep it, we should be extra careful to have just what is needed, otherwise we will waste most of our time manipulating heavy stuff for almost nothing

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* 'N' vs 'Population'
+* ...
+* 'phinull' defined in G_REduced preset 'badnegociation', but not used...
+
+
+Class utility:
+--------------
+* _get_DicOfSameUnits() seems redundant with hub.get_dparam(unit=...)
+* fromOldToNewModelFormalism() seems incomplete and unused 
+* 'func': funcs['logic']: 2 different names for the same thing!
```
DaluS, line 25: I think it's not in the same place ? The formalism for the model is not the fields defined inside the core ( there is a translation into the right field, previously in `fromOldToNewModelFormalism`

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* Preserving _check_dparam because:
+    - more complete checks (better more than less)
+    - preserves functionality of load directly from a saved dict
+
+
+SOLVER:
+-------
+* RK4 seems to be wrong for some reason => to be done later
```
DaluS, line 33: The explanation is simple, two loops has been inverted. 
For the moment the order is : 
* For each variable
* Calculate one step four time, and keep a weighted increment

It should be : 
* Calculate one step for each variable
* Do it four time 
* Do a weighted increment of the four steps

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+
+TESTS:
+------
+* import _core at the top of the file => no
+
+
+LEFT TO DO:
+-----------
+* Cycles analysis
+* Change system size
```
DaluS, line 48: I have done it, I'll add a PR soon

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* => To be done later
+
+TESTS:
+------
+* import _core at the top of the file => no
+
+
+LEFT TO DO:
+-----------
+* Cycles analysis
```
DaluS, line 47: It is working well on `NewBranch` just check that the new core is working well with it !

DaluS (commented):
Here are a few comments that might help on the different observation you've done. Do not hesitate to contact me if some stuff are unclear !

Didou09:
> I saw in your push that you have a "param" explicitly defined by the user in def_fields, are you sure it is a good practice here ? The system automatically recognize what are parameters from the model file in `new_main`, I don't see cases it is necessary to declare it beforehand ? (or when a parameter is a function of parameters ?)

The system also automatically recognizes parameters in this PR.

But it becomes useful to have a `eqtype='param'` for parameters that are themselves _**functions**_ of parameters.
It makes our life easier especially if in the future we have parameters which are functions of parameters which are themselves functions of parameters... (nested functions of several levels).

The user does not necessarily have to explicitly add the `eqtype='param'` field, it will be added automatically if it is missing.
I included it explicitly in _def_fields.py for clarity and as a reminder that it can be done.

notes_Issue87.txt
Didou09: ok

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+* Replaced _DESCRIPTION by __doc__
+* hub.model dict vs str ?
+* __ini__.PrintDFIELDS() and subsequent thnigs (_get_summary()) => To be done later
+* 'i' vs 'inflation'?
+* 'p' vs 'prices'
+* 'K' vs 'Capital'
+* 'D' vs 'Debt'
+* 'N' vs 'Population'
+* ...
+* 'phinull' defined in G_REduced preset 'badnegociation', but not used...
```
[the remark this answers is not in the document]
Didou09: Done in my last commits, thanks to the `eqtype=param` that we discussed above :-)

notes_Issue87.txt
```diff
@@ -0,0 +1,50 @@
...
+
+
+Class utility:
+--------------
+* _get_DicOfSameUnits() seems redundant with hub.get_dparam(unit=...)
```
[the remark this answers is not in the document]
Didou09: ok, will double-check

notes_Issue87.txt
Didou09: ok, has been removed in this PR

notes_Issue87.txt
Didou09: ok, I suggest we keep all checks for now and until we really stabilize (no change of formalism, no change of strategy / architecture). We we have really stabilize and everythong is running fine for months even for the least experienced users then we can decide to get rid of some parts.
As long as we have not fully stabilized and as there may be new contributors, I suggest we keep all checks for safety

notes_Issue87.txt
Didou09: I mean in general it is a risky idea to hae 2 different names for a single thing in the code, wherever that thing is in the code.

In this PR I took the liberty of using the key `'func'` for the functions (see models files), as we did before and as our new contributor did. 
Several reasons:
* It is called 'func' everywhere in the code (consistency + less things to change)
* 'Logic' is already used to name the dict defining the model

notes_Issue87.txt
Didou09: ok, on it today and tomorrow

notes_Issue87.txt
Didou09: Yes I saw, this is just a list of things I saw in your PR that I have not reviewed yet (the reviewing is what is left to do)

DaluS: Ok ! I thought you were thinking about adding the function, it should not be necessary :)

Didou09:
Ok, since our last exchanges this PR has seen +40 commits to try and answer all questions / comments and I think it is ready for a new review and possibly merging.

What's new since the previous discussion (in addition to all that was presented before):
------------------------------------------------------------------------------------------------
* _**`_def_fields.py` and model files:**_
    - cleaner (removed deprecated code in models, moved it to the end of the file for back-up in the short term for _def_fields.py)
    - units can now be at any power (^{-1.5}) as long as it is a int or float power (positive or negative)

* __**class_checks.py:**__
    - properly handles parameters which are defined as functions (e.g.: `phi0` and `phi1`), even the input arguments are themselves functions (in case it happens in the future), see in `_def_fields.py`:
image

    - supports `lambda` functions but also regular `def ` function (we saw in the contribution from @MattArran that users may want to use that when the function is a bit more complex).
      This was tricky:
        - since `def` functions are not one-liners, they cannot be pretty-printed in the `get_summary()` => prints the dependency instead
        - Like `lambda` functions, they cannot be saved, but unlike `lambda` they cannot be re-created from the saved source, so they have to be reloaded from the model file. This is the only limitation: the model file has to be present on the system otherwise saving will work, but loading will fail)
image

    - Since parameters can now functions, a new `eqtype='param'` is introduced to keep track of them and easily update them  if the user changes one of the input fixed parameters
    - The `func_order` is not just a list anymore, it is a `dict`.
      Indeed, both `eqtype='statevar'` and `eqtype='param'` require a function order, stored in `dfunc_order['statevar']` and `dfunc_order['param']` respectively
 
* `Hub.__repr__()`:
   - as requested, now prints also the nb. of parameters (fixed + functions), `ode` and `statevar` + the `preset`, and the source file is moved to the end as it of lesser importance

image

* `Hub.get_summary()`:
    - more complete, it now includes the output from` __repr__()` as a header, and includes the definition as well
    - The issue with the initial values that were not displayed properly is fixed
    
image

* `Hub.get_dparam()` and `Hub.get_dparam_as_reverse_dict()`:
    - `Hub.get_dparam()` is now much more flexible and powerful than it was before, for each quantity, it can handle:
        - a single value (will find maching entries)
        - a list of values (will find entries matching any in the list)
        - a tuple of value (will find entries matching none in the tuple, useful for excluding)
        Keep in mind you can always decide whether to print the output or return it as list or dict
    
image

    - Based on this, an equally powerful `Hub.get_dparam_as_reverse_dict()` method does what was done by `_get_DicOfSameUnits()` (returning a reverse dict) but in a more flexible manner: it can return a dict based on any criterion (not just `units`) and can also take additional selection criteria (for finer selection)
    Here again, you can print or return as a dict
   In the second example below we want parameters (`eqtype=None` or `eqtype='param'`) sorted by `units` (`units` is the sorting criterion with which the dict is created), but only if they belong to the groups `Investment` or `Core`, excluding `k0`:

image

* _**solvers:**_
   - the issue with solvers are not adressed in this PR (will be done a separated PR)
   - However, I noticed that for the crisis scenario, the solvers `eRK4-scipy` and `eRK8-scipy` would not converge after `t=17 s`, This is due to the fact there is a crisis with quantities going to 0 or to infinity. This is new since I implemented `phi0` and `phi1` as functions of `phinull` (did not appear before because I had replaced `phinull` by `phi0` temporarily).
   - While it may be physical (crsis => non-convergence) I implemented a warning so the solver would still give all the results until the crash time (in the previous version if the solver crashed everythong would be lost).

* _**Unit tests:**_
   - more detailed error messages for some tests
   - more complete tests (`Hub.get_dparam()`)
   - A new test (`Hub.get_dparam_as_reverse_dict()`)
   - All passing :-)

* _**Miscellaneous:**_
   - I double-checked that `Hub.FillCycleForAll()` still works fine with this PR
   - Like you had done, I implemented optional messages at instanciation like you had done (they can be disabled by `verb=False`)
   
image

   - Like you had done, I made sure users can define parameters in model files that are not known to `_LIBRARY` (the library is locally automatically updated)

@Dalus, I hope all of this meets your needs, if there is more features / changes to be implemented I suggest we keep them for a later PR unless it is urgent? If you agree, it would allow me to move on and focus more on the RK4 solver issue

DaluS:
Excellent ! Let's use this as a basis, I'll do a list of future possible improvement based on it, and rewrite the tutorials based on this version. 
I just think that `set_dparam` will have a few modifications so that it can take into account parallel systems and array of parameters !

PR #100, opened 2021-08-21
Small Improvements : units, comments... 

DaluS:
Nothing amazing, just a bunch of small modifications : 

* now both `W` and `w` are defined in the library, in accordance to usual conventions. Corrected in models also
* `yr` is used instead of `y` for years


DaluS:
To be done later,

PR #101, opened 2021-08-23
Issue66 packaging first step

Didou09:
Motivation:
---------------
* Make the users' lives easier
* This is the very first necessary step (but not the last) on the road to making the code a proper python package
* This only a bit of code restructuring, that's all

Disclaimer:
---------------
* The large number of files is due to the non-regression tests files
I had to change the path a bit, so git considers I deleted and recreated them.

Main changes:
--------------------
* Created a `gemmes/` directory and put the code in it
* Created a `__init__.py` file in it to make a python package, and to control what's easily visible to the user
* Updated all imports and unit tests => passing
* Tried to update `Cockpit.py`, but you should double-check (in particular at the end not sure which units you wanted to plot with AllPhaseScape)

Typical usage:
-------------------

```
In [1]: import gemmes as gem

In [2]: gem.get_available_models()

In [3]: gem.get_available_solvers()

In [4]: gem.get_available_output()

In [5]: hub = gem.Hub('GK')

In [6]: hub.run()

In [7]: hub.get_summary()

In [8]: dax = hub.plot()

In [9]: hub.save()

In [12]: hub0 = gem.load(<path_to_file_to_be_loaded.npz>)[0]
```

Question:
-------------
* Which name do you prefer ? `gemmes` or `pygemmes` ?
```

import gemmmes as gem

import pygemmes as pygem

import pygemmes as pg

import pygemmes as pgm
```

You have to choose a name (`gemmes` vs `pygemmes`) and also to choose a short name for imports (`gem` vs `pygem` vs `pg` vs `pgm`)

Indeed, the users will use the short name they want, but you just have to make sure you always use the same short name in the documentation / examples they will read. And they will probably naturally adopt the one you suggest in the doc.

Although I did not find official guidance in the PEP8 regarding this point, common pythonic usage for the short name is typically to have 2-3 letters (=> `pygem` is a bit long for a short name), and people typically use consonants only.

Exemples: `numpy as np`, `pandas as pd`, `matplotlib as mpl`, `pyplot as plt`, `scipy as scp`...

Make your choice and I'll update it in this PR (or we can update it in a later PR if you prefer)

Details and explanations:
----------------------------------
* This makes the users' life easier because they don't have to know how the library is structured (they to don't have to know where they can find the main routines, in which module...): they have a single entry point the `gem` package.
* We make the main routines / classes directly available to them from `gem`, it is simpler and shorter
* We choose what is available from `gem` by writing it in the `gem.__init__.py` file
* I already included the most useful parts (the main class `Hub` + `get_available_...()` + `load()`)
* Good practice is to make the minimum number of routines available to not confuse the user with too many things and to keep things simple and clear. This way he/she will see a short list of items easy to understand when he/she will use completion:

image* (but in fact advanced users can still access all routines if they know the internal structure)

* For the same reason, the directories `models`, `plots` and `utilities` have been renamed `_models`, `_plots` and `_utilities`, s they don't show up in the completion.

But of course, you can still have access to them, just add the underscore `_` and they will show up in the completion:

image

* `Cockpit.py` remains at the root path  (i.e. does not go inside the directory `gemmes/`) because it is a high-level interface that uses the library (like a user).

Issues:
----------
* This fixes issue #66 , more issues will have to be opened to address all the steps of making it a proper package


DaluS:
Hi didier ! 

For the questions : 
* I think pygemmes  / pgm is great as a convention 
* I think this wrap is a very good idea
* the allphasespace is depreciated and should be removed for the moment

PR #104, opened 2021-08-25
Issue99 multiple systems

Didou09:
Motivations:
----------------
* We want to be able to run multiple systems in parallel
* By multiple systems we mean: the same model but with different parameters
* It should still be possible to run a single system
* Both single and multiple systems should run with all solvers

Disclaimer:
---------------
* The large number of changed files is due to an update to the non-regression files, necessary for 2 reasons:
     - some of the attributes changed : there is now a `dmisc['dmulti']` sub-dict that hold the info on the number of systems, their shape...
     - I found a small bug that is now fixed: the default value of functions was not always properly set when loading a preset, resulting in wrong results for some non-regression files

Main changes:
--------------------

### What cen be set to multiple values
Up to now, quantities in `Hub.dparam` can be classified into 4 categories:
    - `eqtype = None`: fixed-value parameters
    - `eqtype = 'param'` parameters defined as functions of other parameters
    - `eqtype = 'ode' ` ordinary differential equations
    - `eqtype = 'statevar'` state equations (non-differential equations)
Following this classification, and according to the needs that have been expressed so far, I implemented the possibility of setting multiple values (an array of value) for all fixed-value parameters (`eqtype = None`).
This will be useful for testing the sensitivity of a model to its parameters
However, I have not implemented the possibility of doing the same for initial values of `ode` (partially for lack of time). When we will need to do it, it will have to be done in a separated dedicated PR (they do not belong to the same category as fixed-value parameters and the implementation will have to be a little bit different). Feedback from this implementation on fixed-value parameters will also be useful.

### Setting array of values for fixed-value parameters:
Just use the usual Hub.set_dparam() method, passing an array (or list, or tuple) of values instead of a single value.
You may want to specify a value for grid, which will become relevant when 2 or more parameters have multiple values:
    - grid = False (default): all arrays shall have the same size (= total number of cases), useful for cherry-picking interesting cases
    - grid = True: arrays don not need to have the same size, a grid of all possible combinations if computed. If parameter 1 has n1 values and parameters 2 has n2 values, the results will have shape (nt, n1, n2), where nt is the number of time steps. Useful to sensitivity tests.

With grid=True
image

With grid = False (notice that here both parameters have 5 values)
image

If we try to set only 4 values to the second parameter with grid=False, an Exception is raised with an explicit error message:
image

### Visualizing in terminal:
The Hub.get_summary() method has been adapted to display relevant info even for multiple-valued parameters

With grid=True
image

With grid = False
image

### Solvers adaptation
* The homemade solver is by nature vectorized, so only very small changes where necessary, and the global run still appears as a single run
* The scipy solvers do not allow to solve multiple systems simultaneously, so a loop on the cases had to be implemented, as a result a message is now printed to show the success / failure of each run independently

image

The `Hub.get_summary()` method now also accepts a `idx` argument to specify a particular system to show
    - `grid = False`: `idx` must be a `int`
    - `grid = True`: `idx` must be a `tuple` of `int` indices, each pointing to the corresponding parameter

`grid=False`
image

### unit tests
* A new class of unit tests has been created, dedicated to testing Hub with multiple systems, with and without grid.

Remaining to do in a future PR:
-----------------------------------------
* Implement possibility of setting multiple initial values for ode (high or low priority?)
* Update Hub.plot() with a few more ideas and options with regards to multiple systems

### DIFF
--- a/pygemmes/_utilities/_solvers.py
+++ b/pygemmes/_utilities/_solvers.py
@@ -250,7 +250,6 @@ def get_func_dydt(
         msg = "Vectorized version for implicit solvers not implemented yet"
         raise NotImplementedError(msg)
 
-
     if 'scipy' in solver:
         lode_solve = [k0 for k0 in lode if k0 != 'time']
         shapey = (len(lode_solve),)
@@ -373,7 +372,7 @@ def _eRK4_homemade(
             t0 = _class_checks._print_or_wait(ii=ii, nt=nt, t0=t0, **dverb)
 
         # Estimate dt (for future variable time step versions)
-        # dt = 
+        # dt =
 
         # compute ode variables from ii-1, using solver
         y += _rk4(


Didou09:
> If you ask a grid=True and add two vectors of n=5 elements, are you going to create a matrix (5,5) or just an array (5) ? If you generate a grid first then want to do a modification grid=false after, is it working well ?

In the current version, if you add 2 vector of shape `(5,)` with `grid = True`, you get a `(5, 5)` matrix
If you set `grid=True` first and then change your mind and want to set `grid=False`, you have to reset the vectors (except if they all already have the same size)

> I think we must be cautious with this grid
It is set to False by default,

Only to be used voluntarily when the user wants to explore systematically a space of parameter values (sensitivity studies)

> Indeed it is the regression test that fails, I have to say that it's unexpected... But it goes wrong on a diverging model, so it might explain the origin of the problem. I would accept it that way and maybe upgrade the preset for a "less diverging" scenario

I'll try to fix it in a dedicated PR, it might require changing for a less diverging model indeed, could you suggest parameters values for the preset that would give a less diverging model ?

> I would put it rather high priority, the actual development is great for sensibility analysis, and ODE initial value will be great for basin of attractions.

OK, will try to address this week

Issue #105, opened 2021-08-30
Multiple initial values for ode


Didou09:
Motivations:
--------------
* PR #104 introduced the possibility of setting multiple values to fixed parameters
* It let the user choose whether he wants a systematic exploration of all combinations of values (`grid=True`) or to pick only cases of interest (`grid=False`, default)
* The same would be useful for initial values of `ode`

Proposition:
--------------
* Implement the possibility of setting multiple values for initial values of lde (using arrays)
* It will probably imply small modifications of attribute `Hub.dmisc['dmulti']`
* Make the `grid` option more flexible:
    - now it's either `grid=True` for all or `grid=False` for all
    - It would be nice if the user could choose to set `grid=True` just for some parameters / initial values he picks. This would allow to combine cherry-picking of cases of interest with systematic exploration of some parameters only.

DaluS:
It'd be very practical to have the ODE initial conditions indeed ! 

I think the grid is not too important (compared to multisectorial/EDP) but if it's rapid for you to implement that's a good enhancement

PR #88, opened 2021-08-13
Addition of Putty-Clay model and PDE support

MattArran:
Added a model corresponding to a Goodwin-adjacent model, with a putty-clay production function, from Akerlof and Stiglitz (1969), and changed the code base to support this model. In the paper's model, investors select from technologies with different capital-labor ratios, to maximise profits based on current wages, but the capital-labor ratio is fixed at the point of investment. Consequently, potential employment is distributed at any time over machines of varying capital-labor ratio, with the distribution evolving as a partial differential equation. This model does not yet support multiple parallel systems.

To implement the model, support's been added in `set_dparam`, `reset`, `get_variables_compact`, `get_summary`, and `class_checks` for:
- `np.ndarray` parameters and variables
- multi-line non-lambda `'func'` definitions
- a `'pde`' equation type
These features are not yet supported in analysis and plotting code.

`def_fields` has been updated with the new model's variables and parameters and with corrected dimensions and units (fixing issue #85).

Ref:
- Akerlof, G. A. and Stiglitz, J. E., 1969. 'Capital, Wages and Structural Unemployment', The Economic Journal, Vol. 79, No. 314 (http://www.jstor.org/stable/2230168)

PR #106, opened 2021-09-01
Issue105 multiple initial ode

Didou09:
Motivations:
--------------
* See issue #105 

Disclaimer:
------------
* The large number of changed files is again due to updates to the non-regression files

Main changes:
----------------
* The user can now specify an array of values for the initial value of all `ode` functions
* This is compatible with specifying an array of values for parameters too, both can be done for the same `Hub` instance
* Setting arrays of values be done either dynamically from the command line or written in the preset
* The `grid` keyword argument is not a global attribute anymore, it can now be specified individually for each ode / parameter
* Several hidden bugs were fixed on the way to improve robustness
* New unit tests dedicated to Hub with arrays of parameters / ode initial values, all passing

Detailed changes:
--------------------
* New option: The user can now specify from which dictionnary he/she wants to define the presets (useful for designing custom presets without having to edit the model files + useful for unit tests)
The dict can be passed directly from the command line

image

* The user can set multiple values for one / several parameters and for one / several ode initial values, either using `Hub.set_dparam()` for each quantity or directly inside the preset dict
* When multiple values are provided and `grid` is not specified, it is set to `False` by default and all arrays must have the same size (or an Exception is raised)

image

* The user can decide to explore more systematically the model and set `grid=True` for each quantity indepently (a new dimension is added for each quantity with `grid=True`). This can now be specified in the preset definition too, by providing a dict with keys `'value'` and `'grid'` instead of a just a value:

image

* This works both with the homemade solver and with the scipy solvers

* Several hidden bugs were identified and corrected:
    - the functions were passed by reference, which may pose a problem in the event of a user creating several Hub instances of the same model in the same python session. When he modifies the parameters of one instance (hence the default values of its functions) it would modify the default values of the functions in the other instance too without warning....
    This was **fixed** by systematically copying the function and model loading time (see `_class_checks.copy_func()`)
   - The default values of functions were not properly re-set when an instance was loaded from a previously saved instance file, **fixed**
   - Similarly, when loading from a saved output file, the `Hub.dargs` dict attribute was not passed by reference anymore (because it was populated with saved arrays), thus de-correlating the values in `dparam` from the value in `dargs`, which poses a problem if the user decides to re-run the instance after loading (he will get inconsistent results), **fixed**.
- More detailed error messages were implemented at several places, in particular for `Hub.__eq__()`

* In Pr #104, the non-regression tests did not pass for a large divergence for MacOS only, it may be due to machine precision issues. To try and fix this the possibility of choosen the precision of the non-regression tests has been implemented and a message giving the required precision is now printed if the unit tests don't pass, so we should be able to adjust iteratively, if necessary

* A new unit test file was created, dedicated to testing `Hub` with multiple values for parameters and ode initial values, defined both from the command line and from presets.

Next tasks:
-------------
* Now that we have a stable new formalism with multiple values, I will try to make PR #88 compatible with this new version.
It may take the form of a new PR, and @MattAran, I may need to ask you questions on your needs if there is something unclear to me (although it should not be the case, you code is clean and understandable to me as far as I looked at it). If you agree @MattAran, I will put you as reviewer of the corresponding PR so you can double-check that your needs are met, as well as @Dalus as usual.


Didou09:
Non-regression:
------------------
* All unit tests (including non-regression) are passing on my laptop
* Again, one non-regression (the G_Reduced_badnegociation_eRK8-scipy case) is not passing on Github, this time also on linux
* The more detailed error message implemented show that:
    - it concerns between 1 and 3 values out of an array of 10000 values
 vs     - The difference with respect to the reference file is typically of the same order of magnitude has the numerical tolerance (for example 7.027291516470413e-07 vs 1.0909310318119782e-07)

image

This tends to confirm that:
* we should set a less diverging case for non-regression tests (less sensitive to numerical errors)
* and / or we can increase the numerical tolerance for non-regression

I think the first option is better-practice, and the second option should be used only if it does not work, but that's only my opinion and I can understand the second option may be more practical on the short term.

What do think @Dalus ?

I all case, I suggest we fix thi before merging, merging a PR with tests failing is probably the worst option...

DaluS:
Thanks for the PR ! 
I did not test this version for the moment, I'll do it in the following day. 

My suggestion for the regression test : If the quantity `x` is diverging, testing it on `x^{-1}` would do the work : either one or the other has to be close enough

Issue #107, opened 2021-09-01
Code clean-up and simplifications


Didou09:
Motivations:
--------------
* We now have a running code, and it is slowly getting more and more mature
* As it matures, we can start identifying small changes that could make it easier to maintain / develop, or that could simplify it or make it more straightforward
* Based on my development experience from the last weeks, this a small non-exhaustive list of small changes that would out life easier at developping / maintaining.
* I will gradually list here all small changes I identify as I code
* Please comment below so we can decide together

Reserved keywords avoidance:
----------------------------------

### lambda
`lambda` is a common variable in macro-economic models
It is also a reserved name in python (for anonymous functions)

When the user defines an anonymous function that depends on `lambda`, there is a name conflict, it is not possible to write

![image](https://user-images.githubusercontent.com/5929562/131649622-3c2d509f-53c9-478b-a74c-3d4f90320fcb.png)

So far, this has been circumvented by using lamb instead of lambda in such case (in the keywaord args of functions.
It works, but it introduces significant hassle because it forces the developper to always treat lmabda as an exception and to implement if statements to treat it separately

This has slowed down development and made it more difficult (if you forget to treat that exception the code fails and you have to understand why).

A very simple fix that would simplify the code and development would be to uniformize the keyword argument and the dict key as lamb, so routines can treat it like any other variable and automatically associate the keyword arg with the dict key.
The full name lambdac an still be stored under the symbol field, which is to be used for plots, this way plots would still display lambda, but the developer's life (and the code) would be made non-negligibly simpler.

Replace this:
![image](https://user-images.githubusercontent.com/5929562/131650676-39fdb41c-c8c5-4600-958b-d0a1acf0a9de.png)

by this:
![image](https://user-images.githubusercontent.com/5929562/131650591-29ca75b7-ede9-48ff-b05c-f43e82cd523e.png)

File / models naming:
-----------------------
* According to PEP8 conventions, most variable and function names in python have to be minor case and words separated by underscores
* In the same spirit, we are currently using underscores to separate the various parts of an output file name (and even of module names)
* This allows for easy parsing of the file names to extract the relevant info
* But it only works fine if each part of the name does not itself have an underscore
* For example, the non-regression files are stored as:
![image](https://user-images.githubusercontent.com/5929562/131655374-eb5ccd43-d139-437c-884a-08813f85904e.png)

This was already discussed in PR #62: the saved files have structure:

`Output_model_preset_solver_<custom-name-from-user>_user_date_time.npz`

[...]

PR #112, opened 2021-10-21
Clean name/dimensionless

DaluS:
# Update ! 

I'm baaaaaack, sorry for the delay. Here is a list of changes : 

## Fields file

* I reorganized fields in a way that seems more practical
* Verification of units 
* dimensionless units are now explicitly taken into account through units = '' ( None mean that it is not defined )
* There is a temporary section on the top for fields that are not definitive and introduced in models, as I struggle with adding them inside the model file (the system try to put them in the resolution order of the equations)

## Models 

* There is still a small number of models ( GK-reduced and GK ) for economy because I'm still struggling with the solvers
* I added non-economic models ( damped oscillators and lorenz system) for general testing 

## Multiple cockpit files  

* `TEST-CONVERGENCE` test the different solvers both quantitatively (on the damped oscillator) and qualitatively (on Lorenz)
* `BASIN-ATTRACTION-GK` use the potential of parallel system solving to look at the relation between initial and final value. It calculates the boundary between good and bad attractor in a GK model given (lambda,omega,d) when t=0

mygif

_the GIF : the color of each pixel is private debt value in time, the position of each pixel is the initial lambda/omega value. When pixel stay red, the trajectory is going to bad equilibrium. The shape of this frontier is found for each  initial value.  This has been calculated on runs with the homemade rk4, some runs are not working hence the white square. This gif thus contains 20*20*11 simulations with 20/0.005 timesteps, with one minute total computation). Ultimately the shape of the basin of attraction can be calculated using each layer of initial debt value, showing this type of lines.._

image

## What are the next things ? 

* Understanding why the RK8 Solver sometimes fail in case it should not 
* Understanding why the homemade RK4 is not working the intended way
* Include elements developed in the different cockpits files into pygemmes files
* Write correct presets so that unit test are on preset values
* Add more models of course 
* Review of the plots section to allow same units variables to be plot altogether
* Allowing multisectorial writing (AKA Matt's development)


DaluS:
It's working well on my computer, I have absolutely no idea why these tests fail on github, can @Didou09 you check on your side ?

DaluS:
The error comes from test multiple which was tried on a model name written explicitely, that had been changed previously. Next pr #116 solves it

pygemmes/tests/test_01_Hub.py
```diff
@@ -29,6 +28,7 @@
 
 # library-specific
 sys.path.insert(0, _PATH_PCK)   # ensure Main comes from .. => add PYTHONPATH
+import pygemmes as pgm
```
Didou09, line 31: This was already done in devel, are you sure you started from devel ?

pygemmes/_models/_model_GK-Reduced.py
```diff
@@ -34,9 +34,8 @@
             'com': 'reduced 2variables dynamical expression'
         },
         'omega': {
-            'func': lambda itself=0, phillips=0, i=0, gamma=0: itself * phillips - (0.5)*i,
+            'func': lambda itself=0, phillips=0, i=0, gammai=0, alpha=0: itself * (phillips - (1 - gammai)*i-alpha),
```
Didou09, line 37: if gammai does not match a variable name the code won't be able to reckognize it in the dependence-searching

pygemmes/tests/test_01_Hub.py
```diff
@@ -28,7 +29,6 @@
 
 # library-specific
 sys.path.insert(0, _PATH_PCK)   # ensure Main comes from .. => add PYTHONPATH
-import pygemmes as pgm
```
Didou09, line 31: why remove it ? This part was to make sure the local version was loaded

Didou09 (commented):
Commit 1:
---------------

* By convention, since I only have limited time, I only check what is inside folder `pygemmes/` (+ `setup.py` and `requirements.txt` for the packaging), so I won't check changes to `Cockpit.py` and all other files / folders not in `pygemmes/`. 
This way @DaluS you can do what best suits your needs outside of `pygemmes/` without me interfering, and I focus on the cor part of the package.

* I agree with the name changing of `_model_G_Reduced.py` to `_model_G-Reduced.py`, this is in accordance with the conventions discussed in issue #107 

* `_models/_def_fields.py`:  ok, i understand you're setting default units to '' instead of None and introducing 2 new parameters

* `_models/_model_GK.py` : ok, I understand you're introducing more complete functions for `phillips` and `I` and introducing new function `solvability`

I am ok with this first commit

Didou09 (commented):
Commit 2:
--------------

* Overlooking `Cockpit.py`

* `_models/_def_fields.py`: ok, introducing new parameter `gamma`

* `_models/_model_GK-Reduced.py`: ok, introducing a brand new model

* `_models/_model_GK.py`: ok, making several functions more complete

I am ok with this commit

Didou09 (commented):
Commit 3:
--------------

* Overlooking `Cockpit.py`

* `_core.Hub. FillCycles()`: ok, removing useless comment block + adding debugging print

* `_def_fields.py`: ok, replacing `zphi` by `zsolv` (to avoid naming conflicts ?), and adding `solvability`

* `_model_GK-Reduced-SolvProfit.py` : ok, introducing brand new model

* `_model_GK-Reduced.py`: ok, replacing `zkappa` by `zsolv`, adding solvability

* _model_GK.py : ok, replacing `zkappa` by `zsolv`


I am ok with this commit

Didou09 (commented):
Commit 4:   #b8ac5acd77195cba0c6717b46bbdd439a3a71146
--------------

* Overlooking `Cockpit.py`

I am ok with this commit

DaluS:
> * By convention, since I only have limited time, I only check what is inside folder `pygemmes/` (+ `setup.py` and `requirements.txt` for the packaging), so I won't check changes to `Cockpit.py` and all other files / folders not in `pygemmes/`.
>   This way @DaluS you can do what best suits your needs outside of `pygemmes/` without me interfering, and I focus on the cor part of the package.

Yes ! I think that's the best way ! 
There are "optional" libraries (to draw variable graph/basin of attraction), that we should add (the necessary libraries are the same as before)

Didou09 (commented):
Commit 5: #0859e0fefd59a5476f8850fcd3120cb38e054c50
--------------

* Overlooking  `Cockpit.py`

* `_def_fields.py`: ok
    * passing `time` into `Numerical` group
    * Renaming `CORE` group to `Household`
    * Removing `beta`, `delta`, `omega`
    * Reorganising several groups of parameters
    * New `Consumption` group
    * Introducing new parameters `gammai`, `GDP`, `Recipes`

* `_model-MonoGEM.py`: not ok, brand new model BUT, the file name is not conform to the accepted conventions ! => will be corrected in future commits?

* `_model_GK.py`: ok, changing a bit equation on `W`

* `_model-MonoGEM.py`: ok, brand new model


I am ok with this commit, provided the wrong file is removed in future commits

Didou09 (commented):
Commit 6 :   #e88f27dcdaf88c8170e15c2ab0ad0520cb56e723
----------------

* Overlooking `Cockpit.py` and `Cockpit2.py`

* `_def_fields.py`: ok, minor changes in the parameter library

* `_model_CES-Reduced.py`: ok, brand new model

* `_model_DampOscillator`.py: ok, brand new model

* `_model_GK-Reduced.py`: ok, minor changes to existing model

* `_model_LorenzSystem.py`: ok, brand new model

* `_model_MonoGEM.py`: ok, removed model

* `_utilities/_class_checks.py`: ok, minor indentation change

I am ok with this commit

Didou09 (commented):
Commit 7:     #1e99d6d84702d010b3e88d5740cd144eb04c965d
---------------

* Overlooking `Cockpit2.py`

I am ok with this commit

Didou09 (commented):
Commit 8: #a950efca3019278562ee39e78df5f0343d234c6e
--------------

* Overlooking `BASIN-ATTRACTION-GK.py`,  `Cockpit.py`, `NotIncluded/Cockpit2.py`, `NotIncluded/Cockpit3GK.py`, `TEST-CONVERGENCE.py`

* Several models moved to `NotIncluded/`: ok

* `NotIncluded/test_02_Hub_Multiple.py`: not ok a priori, this means we are removing a previously working unit test, which is bad practice but can be justified if temporary. Will depend on future commits

* Several models modified: ok (`DampOscillator`, `GK-Reduced`, `GK`, `LorenzSystem`)

* Non-regression reference test files removed: not ok a priori, depends whether they are properly replaced by updated ones in later commits

* `pygemmes/tests/test_01_Hub.py`: not ok
    * import pygemmes moved to top: not ok because it cannot ensure that the local version is loaded (will have to be corrected in future commits)
    * Removed replacement of underscore by dash in model name: not ok because it forces conventions, will have to be corrected in future commits

I am ok with this commit, with the following caveats that have to be corrected in future commits:
* `test_02_Hub_Multiple.py` is necessary
* Non-regression tests are necessary
* `test_01_Hub.py` will have to be corrected backwards

Didou09 (commented):
Commit 9 (last):   #3c0b2e717a780af44d63ef099f080e0e519c7c64
---------------------

* Overlooking `Cockpit.py`

* `NotIncluded/_model_GK.py`: ok, but may be necessary for unit tests ?

* `_model_GK-Reduced.py`: ok, minor change

* `test_01_Hub.py`: ok, import pygemmes bacj in right place (remove one caveat from previous commit)

I am ok with this commit, but there are still 2 remaining caveats: `test_02_Hub_Multiple.py` and non-regression tests.

+ Unit tests are not passing, so with your approval @DaluS I will fix unit tests and solve the 2 remaining caveats

DaluS:
> All susequent commits were validated in PR #124

I close this PR in consequence

Issue #113, opened 2021-11-17
Solver RK4 still surprising behavior


DaluS:
I had trouble with the homemade rk4, and decided to verify if our solvers were working well. In consequence, I coded a very simple couple ODE (dampened harmonic oscillator) and compared results with its analytical solution. 

The Homemade RK4 still have a weird behavior. Here is a difference between the numeric and theoric value : 

![image](https://user-images.githubusercontent.com/11523050/142274450-fab6e19b-f9a1-4fe8-80bf-4aae8921ad3c.png)

The slope is 1/2, it should be 4, this system surprisingly behave as a system of order 1/2.

![image](https://user-images.githubusercontent.com/11523050/142275146-117a1c43-7991-4569-9619-1abd88dcbd6b.png)

It seems to me that the code is right, so I do not know why it is not working https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods

```
def _rk4(dydt_func=None, dt=None, y=None, t=None):
    """
    a traditional RK4 scheme, with:
        - y = array of all variables (all ode)
        - dt = fixed time step
    """
    dy1_on_dt = dydt_func(t, y)
    dy2_on_dt = dydt_func(t + dt/2., y + dy1_on_dt * dt/2.)
    dy3_on_dt = dydt_func(t + dt/2., y + dy2_on_dt * dt/2.)
    dy4_on_dt = dydt_func(t + dt, y + dy3_on_dt*dt)
    return (dy1_on_dt + 2*dy2_on_dt + 2*dy3_on_dt + dy4_on_dt) * dt/6.
```

Didou09:
Will try to investigate this in a few weeks (mid-December probably) if the problem is still there by then :-)

DaluS:
Yes ! I'll focus on using the rk8 from scipy for the moment :)

DaluS:
I coded a Lorenz system to check the  qualitative system, for the moment both the rk2-scipy and the homemade rk4 are wrong :) 
![image](https://user-images.githubusercontent.com/11523050/142455778-57a96859-84db-4951-870a-6d767c3cb377.png)

DaluS:
I went a bit further on the subject : 
* an Rk1 has been added for the very-simple case 
* Coded a convergence test (see `testConvergence_DampOsc` and both Rk1/Rk4 are not power-law which imply that either the convergence test is wrong or that the problem is somewhere else 
[...]

PR #116, opened 2021-12-03
Extensive models & Wrapup

DaluS:
# The big PR 

This Pull Request will contains most of last month advances, since I was alone on this segment, and since it is composed of increments here and there, they are all together. Little has been done on the core itself (which I have difficulties understanding and modifying), So it's a sum of smaller contributions. 

This PR is based on #112 

# ISSUES

A lot of small (or less small) issues still exist in the core as it is :

Here is a list of issues that has been observed : 

## Python routing

Some elements are generated the first time a terminal is 

## Model files

1. We cannot define parameters inside the model file itself 
2. `time` cannot be called using scipy solvers #115
3. plot in preset is still not implemented
4. There is no reason for `_FUNC_ORDER` to be in the file as it is overriden anyway 

## Hub load and presets

1. When creating a hub with a model and a preset, the model is kinda loaded twice with something unclean on undeclared parameters _However the preset is working as intended_
image
2. `set_dparam` works well for one value but erase everything in the case of a dictionnary
image
image
3. Iterating over a dictionary works almost but for grid which is written with another formalism in both cases ( if in a dictionary `("key" : {'value':vector,'grid':True})`, or in a unique parameter `(key="key" ,value=vector,grid=True)`,
4. dpreset works well ! Day-to-day use is now creating a dictionary, putting it inside dpreset with a name, and creating the hub with dpreset and the name of the preset.  

## Solvers 

5. I have found no way to use scipy solvers on constant time-step, making it difficult to verify that the way we write our system is indeed working...
6. The Rk4 does not behave as a Rk4 ? It is unclear
7. Scipy solvers are not functioning in parallel

## Plots 

This is clearly minor ! 

8. Linestyle nor linewidth taken into account in hub.plot
9. `hub.plot` just to plot is not 

## Misc

* The documentation is out of date (doc folder, readme)
* Requirements should be updated too

# Modifications 

## Small modifications 

* Now the structure has a folder "example" that contains cockpit-style files. It accelerate the use of the library in general. `AllFormulations.py` contains all the commands with typical use 
* Lot of files have been displaced to the NotIncluded Folder, and should not be reviewed
* Def fields have now carbon concentration as a unit `C`, a `temporary` group in the beginning of `_LIBRARY` because of specific parameters to one model
*  `` means dimensionless, and `None` is the default (easier to code that way given the existing structure)
* `n` has the role `beta` had before in goodwin-keen type of models (population growth), to allow `beta` being something else
[...]

PR #130, opened 2022-01-06
Issue113 solver issues

Didou09:
Motivations:
----------------
* The homemade RK solvers were showing strange behaviour (see discussion in PR #95 and issue #113).
*  Also, although the RK4 scipy solver was apparently fine, it was surprising that the RK8 from scipy seemed to less stable than the RK4
* A first explanation regarding the homemade solvers was the time step size. Smaller time steps improved the behaviour. But various tests in very simple cases (DampOscillator) coded by me, @Dalus and @timongo raised concerns that there must be something else wrong

=> These solver issues are finally fixed in this PR

Main changes:
--------------------
* The fact that almost all solvers seemed to have an issue pointed to the suspicion that the problem may not lie with the solver themsselves but rather with a a fraction of code that they had in common
* In order to allow the user to freely choose the solver of choice between one homemade and 3 scipy solvers, and in order to avoid code repetition (which is a source of errors) and to ensure common behaviour whatever solver is chosen, the code is structured such that:
    * The model is put into the form of a single dy_on_dt function, aggregating all ode's and computing intermediate state variable for each time step if necessary, this aggregated derivative function is common to all solvers and coded before the solver is chosen, in `_solver.get_func_dydt()`, which return `func` (= dy_on_dt)
    * In order to avoid creating a large amount of intermediate arrays each time this function is called by a solver, I implemented a buffer system so the intermediate computations always overwrite a temporary buffer, passed by reference
    * But I had not paid attention that this buffer is also passed by reference as an **_output_** of `dy_on_dt`. This is what caused the issue common to all solvers: they would call dy_on_dt multiple times for each time step (4 times for a RK4 solver), and each time the previous value would be overwritten resulting, for a RK4 solver into 4 times the exact same term.

* While this was a bit hard to understand and find, it  was very easy to fix: the function `dy_on_dt` now returns a copy of the buffer ad not the buffer itself.

* As a result, this fix also simplifies the non-regression tests failures that we observed in PR #116, [here](https://github.com/DaluS/GEMMES/pull/116#issuecomment-1005620605): now only the Lorenz attractor needs to be tolerated, the DampOscillator does not fail anymore, which is more logical :-)

* Other minor changes include small fixes and improvement to the expression of the Energy in DampOscillator, and comments therein.

* Non-regression tests data is updated according to this new, fixed, version of the code

Issues:
----------
* This fixes issue #113 

@Dalus, I won't be available tomorrow, so please don't hesitate to take the time to re-run all the routines you coded to double-check the solvers again, including basin of attraction, Lorenz attractor... 
You might want to re-run your sensitivity studies too, and redo some figures just to be sure...
Also don't hesitate to do a full review of this PR, it is very small since the fix was very easy to implement (there are only 2 non-binary files modified)

@timongo, thanks for you quick help this evening, it confirmed what @Dalus and I had observed: the solver itself was fine, the truth was somewhere else  ;-)  Don't hesitate to review the PR too if you wish to

Example:
-------------
* Using again the example I gave in issue #113, we now see that:
    * The energy is quasi-constant for a perfect oscillator, as it should (thanks @timongo for pointing this error out)
    * The Homemade RK4 is now quite relevant and robust, and dt = 0.01 is sufficient to give very satisfactory results
    * The scipy RK4 is good and the scipy RK8 is even better

(N.B.: eRK8-scipy and eRK4-homemade with dt = 0.01 y are superimposed)
image

Zoom to distinguish eRK8-scipy and eRK4-homemade with dt = 0.01
image


pygemmes/_utilities/_solvers.py
```diff
@@ -394,7 +394,7 @@ def _eRK4_homemade(
             dydt_func=dydt_func,
             dt=dparam['dt']['value'],
             y=y,
-            t=np.nan,
+            t=np.nan,       # no model with explicit time dependence (for now)
```
DaluS, line 397: There will be soon some ! pseudotime is a good workaround for the moment

DaluS:
Excellent ! I could not have found it clearly :D Thank both of you @timongo @Didou09 ! 

I'm OK with all the changes, and all test I've run this morning show much more stable resolution. That's a huge improvement, stability/sensitivity analysis and frequencies calculation are much closer to what I find in literature ! 

That's a huge weight that disappear, now we can fully focus on the results we extract from our models.

Issue #134, opened 2022-01-07
Article figures reproduction


Didou09:
Motivations:
----------------
* As pointed out already in several issues, and in PR #129, one of the prerequisites of pygemmes is to be able to reproduce some classic figures from various chosen papers
* Most of the preparatory work has been done already by implementing the models formalisms and the presets.

Proposition:
---------------
* In the end, pygemmes should be able to reproduce, on-demand a large number of papers and figures
* In order to prepare for this and make it easily accessible to the user, I suggest to implement a formalism similar to models / presets and building on it
* I am only adressing the core code structure to make that easy to read / execute, not the models themselves.

In the same spirit of what we've done already, I propose to implement something like:

A tool similar to get_available_models() and get_available_solvers()... that will  print a list of the articles figures that can easily be reproduced by pygemmes
![image](https://user-images.githubusercontent.com/5929562/148555738-4378f24c-e27b-45a9-92b5-a72432b4dca5.png)

A single function call that will actually run the reproduction and print the desired figures and compare them with a png of the original
![image](https://user-images.githubusercontent.com/5929562/148556043-1f3c965c-0acc-4d9c-b97a-c1e8f175862c.png)

All the complexity is already stored in the model files and preset (@guittcam 's work if I understood correctly), that this routine will call.
The plotting routines can be developped in a dedicated Cockpit until they stabilized and validated and included directly in the core.

I will propose today a first implementation that we can discuss afterwards,
I can also propose to assist @guittcam on the implementaion in necessary

DaluS:
I like the idea !

An article has often few (2-3) models inside it with each a few (2-3 again) figures. There should in consequence be a dictionary "articles" with the following architectures : 
articles = {1 name_of_the_article : {2 name_of_the_model : {3 name_of_the_preset ,} } }

I think this step will not be too harsh, as long as the "plot" section of presets is efficient. And for the moment I do not know which formalism to use (see #118 )

Didou09:
Fixed, see PR #136

PR #136, opened 2022-01-10
Issue134 articles reproduction

Didou09:
Motivations:
--------------
* Provide a template for implementing articles reproduction
* Implement the user-oriented tools for listing available articles (based on get_available_model()) and for reproducing the article desired.
* see issue #134

Main changes:
----------------
* [NEW] The user can now choose wether he/she wants to load a model from the library's default models form his/her own `$HOME/.pygemmes` customized folder
* [NEW] `pgm.get_available_articles()` is implemented, to list all articles figures available for reproduction, also prints the model and preset corresponding to each figure, and a short version of th article's reference
* [NEW] `pgm.reproduce_article(article=..., figure=..., save=...)` is implemented to reproduce on demand any available article figure
    * optionally the reproduced figures can be saved (`False` by default)
    * The original figure is automatically plotted alongside the reproduced figure, for comparison
* [NEW] Unit tests are implemented to test the reproduction of all available articles
   _**N.B:**_ they are unit tests (they make sure the routines work), they are not  non-regression tests (they do not check that the figures are properly reproduced, i.e.: they check a result a produced without bug, they don't check the content of that result.
   When we have implemented a first real article (not just a template), it will be worth implementing a non-regression tests to that new article, to make sure all subsequent version of the code actually really reproduce the corresponding figures.

Implementation details:
--------------------------
* A new `pygemmes/_articles/` folder is created
* For each article to be added, a dedicated `ARTICLEKEY/` folder must be created
* In this `ARTICLEKEY/` folder:
     * a `_article.py` module has to be created (contains the details on how to create each figure)
     * a separate `figXX_name.png` file containing the original figure must be stored.
       By default, `XX` is the number of the figure (in the original article), and `_name` is a custom name that can be added to easily remember what is plotted on the figure (this part of the file's name won't be used by the code, this is purely for reminding us what the figure is about).
* I have create a template article folder containing all necessary elements, including a full and working `_article.py` file, don't hesitate to copy-paste the whole folder to implement a new article.

Recommendations for adding a new article:
------------------------------------------------
1. Open _**a specific issue**_, dedicated to adding this particular article (i.e.: good practice is **_one dedicated PR_** for each new article)
    In the dedicated issue, specify which reference exactly you want to add, which figures.
    The issue can be used by others for discussing the relevance of each figure choice, for argumenting for / against each choice.
2. Once the choices are made, if the corresponding `models` / `presets` are not yet available, implement them (in `pygemmes/_models/`)
3. Copy-paste the folder `pygemmes/_articles/ModelExampleTemplate/` and rename it to a unique key that will identifiy the new article
4. Edit `_articles.py` and replace all `png` files by captures of the relevant original figures
5. Test, and if necessary ask for help

Issues:
--------
* Fixes issue #134 

Examples:
-----------

To load a model from your private $HOME/.pygemmes/ (default):
image

To load a model from the library's public models:
image

Articles
::::::::::::::::

To get the list of available articles
image

An article is identified by a unique key, here `ModelExampleTemplate`.
An article can implement several models and for each model several presets (i.e.: several parameterizations of the same model).
It is assumed that each figure we want to reproduce can have its unique (`model`, `preset`).
The code automatically groups all figures with the same (`model` / `preset`) to run each (`model`, `preset`) only once even if it corresponds to several figures.

The figure numbers correspond to the figure number _**in the reference article**_.
In the above example, we see we can reproduce figures `2`, `3`, `5` and `6` from article _"Great author, Coping 2015, A very famous journal"_.

To actually reproduce the figures:
image

If we had not specified the figures we wanted, by default all figures would have be reproduced.
The original figure (stored as a `png` in the article's folder) is automaticallt plotted on the right side of the reproduced figure, for direct comparison.

image

image


DaluS:
Great complement / new paths, I merge !

Issue #140, opened 2022-01-14
Pytest does not find pygemmes on branch ShouldMerge


Didou09:

DaluS:
Done a library was loaded as `pygemmes.utils` and not `.utils`

Issue #139, opened 2022-01-14
Pytests stuck because loads users' private models!


Didou09:
Solved by PR #141

PR #141, opened 2022-01-14
Issue139 tests load users models

Didou09:
Motivations:
---------------
* Some unit tests were running from models found (or no found) in the user's private `.pygemmes` repo
* Article reproduction was not always loading from pygemmes either

Main changes:
--------------------
* All fixed thanks to keyword `from_user=False`
* All unit tests passing on my PC with an empty private `.pygemme/_models/` repo

=> @DaluS , once you merge, can you update your local devel branch and run a test on your pc for cross-checking ?
=> if it works, can you then update your local ShouldMerge branch and try again ?

Issues:
----------
* Should fix issue #139
* Could maybe magically fix issue #140 too perhaps ?

### DIFF
--- a/pygemmes/_articles/_reproduce.py
+++ b/pygemmes/_articles/_reproduce.py
@@ -230,7 +230,13 @@ def reproduce_article(
             continue
 
         # load and run
-        hub = _core.Hub(model=v0['model'], preset=v0['preset'], verb=verb)
+        hub = _core.Hub(
+            model=v0['model'],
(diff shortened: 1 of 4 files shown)


DaluS:
Still something strange on my side : `test_01` is working but not the other ones... because he is in the file while not finding them ?
image

DaluS:
It's not resolving the problem on my branch `ShouldMerge` Either 
image

PR #143, opened 2022-01-21
Having parameters in a model

DaluS:
It was stupidly simple however hidden...Now you can add parameters, just with : 

```
    'param':{
        'test': {
            'value':1,
            'com': 'hi I am a test',
            'definition': 'I want to appear',},
        },
}
```

for example after statevar dic in `_LOGICS`


DaluS:
@guittcam can you try with one of your model file ? To do so : 
* Get all your modifications of def fields in a temporary file 
* Save your files models on your disk (not in Pygemmes)
* Go on this branch 
* Load any model in the hub , it will give you a path to a folder you can put all your personal models in
* Put the model file and the parameters as above
* Try to load it

This way you'll also work with the latest version ! Plenty of new functionalities as describe in `doc/tutorial.py`

guittcam:
Reviewed ok to merge

PR #144, opened 2022-01-24
Toward V1.0 : `_def_functions`, `_plot_preset`, `CES`...

DaluS:
# Summary : 

1. Readme now have an explanation on how to contribute
2. `_models` now have a a `_def_functions` file, and an exploration functions `pgm.get_available_functions` linked to it
3. New plots have been created and are accessible in presets
4. CES explicit functions has been written in `_def_func`
5. set_dparam is now working 

# Set dparam works again

The problem was that `set_key_value` was changing the value of `ODE` which is then reset by the system. 
I simply modified the key to change 
image

# New functions ! 

**Motivation** : A lot of functions are shared from models to models, and some models are simply one field to be different. It is easier to write files, compare them and modify them if they have the same bricks. 

**How is the file structured**: The file is an ensemble of class, each having a docstring and functions inside. Each class is a general category (population, productivity...) and contains logic elements. For example : 

image

The formalism is the same as in a model file ! 

Each function is implicitly associated with a specific field in mind ( it is not explicit for the moment).
Sometimes there can be two fields in a same class, as they are linked in their definitions. 

image

Some more elaborate case exist. For example, a production function is often associated with optimisation of profit : the expression of L is deduced from the expression of Y which in term of equations might be difficult. 
I created in consequence subclasses, each fields should be loaded together in consequence : 

image

It can be read through `pgm.get_available_functions()`

image

The files now read as : 
image
 

instead of 
image

Is it better ? I'd love to discuss it with you all. 

**Improvements :**
* `itself` should not be used at that type of place
* The name of fields it should be applied on should be more explicit ? 
* There should be a field in `pgm.get_available_functions()` giving which models use it (same as in `get_available_fields`) 
image
* They are lambda function because it allows better readability ! functions are possible of course.

# Plots 

## New plots 

You can plot all the variables sharing the same units : ( plot name : `plotbyunits` ). Not much control on this.
image

You can plot 3D with colors using `plot3D` with the following controls `_plot3D(hub, x, y, z, cinf, cmap='jet', index=0, title='')`, `x,y,z,cinf` are name of fields.

image

Same idea with phasespace ` x='omega', y='lambda', color='time',`

image

You can plot as many y axis on the same figures if you want to be concise in term of figure density : 

image

'x': 'time',
 'y': [['lambda', 'omega'],
       ['d'],
       ['kappa', 'pi'],
       ],
`_plotnyaxis(hub, x='time', y=[[]], idx=0, title='', lw=1)`

## Calling these plots in preset

A new formalism has been introduced to associate figures and preset : 

```
         'plots': {
            'timetrace': [],
            'plotnyaxis': [{'x': 'time',
                           'y': [['lambda', 'omega'],
                                 ['d'],
                                 ['kappa', 'pi'],
                                 ],
                            'idx':0,
                            'title':'',
                            'lw':1}],
            'phasespace': [{'x': 'lambda',
                            'y': 'omega',
                            'color': 'd',
                            'idx': 0}],
            'plot3D': [{'x': 'lambda',
                        'y': 'omega',
                        'z': 'd',
                        'cinf': 'pi',
                        'cmap': 'jet',
                        'index': 0,
                        'title': ''}],
            'plotbyunits': [{'title': '',
                             'lw': 1,       # optional
                             'idx': 0,      # optional
                             'color': 'k'},  # optional
                            ],
```

The structure : 
it is a dict, each key is a plot in `_def_plot`, and the value is a list. 
each ingredient of the list is a dictionary, that contains all the arguments to pass to the plot function. 

**To use it** 

If you have your preset loaded : `hub.plot_preset()` after a run
if you want to use the preset plots without the preset (it might not work if you need multiple indexes in the preset ) `hub.plot_preset(preset='default')`

## CES 

The extensive file `GK-CES` is ready. 
There are many ways to write the production function, I tried to show the different characteristics that I found interesting under the spotlight ! 
`pgm._core._plotnyaxis(hub,x='time',y=[['l'],['omega','lambda'],['d','nu']])`
image

# Next step

- [ ] Finishing the file reduced equations
- [ ] Verify that they give the same dynamics
- [ ] Reproduce all graphs from the article\


DaluS:
@guittcam @Didou09 if you can check this new version, I'd love to have your feedback ! 
I'm working on my models in the meantime

guittcam:
The new function formalism is probably better because 
- it makes the file very clean
- it reduces the risk of typo causing errors

DaluS:
Can you test that :
* Your code is working with it (especially `set_dparam` which has a weird behavior)
* You can use all the new functionalities !

DaluS:
Added `doc\BasinExploration.py`, a few elements to explore basin of attractions in a more automated way

Didou09:
All these new functionalities are very cool !
Congratulations and sorry for the long silence, these are busy times,

Agree with @guittcam , the new function formalism is cleaner and less prone to typos.

I'll try to come back gradually in the coming week and do some tests and clean-up in the unit tests (and come back to work on the distribution topic).

PR #149, opened 2022-02-05
A lot of models

DaluS:
So that's an exploration of small elements on the themes of GK. No real feature in the core, but a lot on the side. 

The big things are on the model file side : you can call a model and modify it on the fly ! It becomes much easier to understand what you are doing, as you are constructing on something solid and checked. 

A small example : `Coping with collapse` (AKA GEMMES) is a Goodwin-Keen model, with a 3 layer climate model, and some glue in the middle. We can in consequence call the two models, and put the glue. 

How does it works : the model begin as : 
```
# ----------------------------------------------------------------------------
# We simply do a few modifications on two previous models : we load them as a basis
from pygemmes._models import Funcs
from pygemmes._models._model_GK import _LOGICS as _LOGICS0
from pygemmes._models._model_Climate3Layers import _LOGICS as _LOGICSClimate
from copy import deepcopy

# The model is a start of a Goodwin-Keen model
_LOGICS = deepcopy(_LOGICS0)

# We add the climate model sector
for category, dic in _LOGICSClimate.items():
    for k, v in dic.items():
        _LOGICS[category][k] = v
``` 

Then we do our new logics (that can overwrite the previous one)
```
_COUPLINGLOGICS = { \... }
# We add the coupling
for category, dic in _COUPLINGLOGICS.items():
    for k, v in dic.items():
        _LOGICS[category][k] = v
```

And TADA ! Your model is done. Fast prototyping, with a common language. 
Each model should be autonomous. But as `dfields` has default values connexions with exterior models will simply be exogenous constant.

* A doc file has been added `Lotofmodels` that test a few models 
* Useless model files has been removed 
* Class check has been lifted to authorize a few things it should have authorized before


DaluS:
@Didou09 test work fine on my computer, but they take hell of a time... so much that Github stop them.

PR #154, opened 2022-02-18
Camille test02

guittcam:
Minor changes in order to add a _model with a CES production function instead of Leontief optimization.

- created a new model corresponding to the [Green Growth article](https://www.overleaf.com/read/jzmpjddqdchd): `_model_CamilleFullCES` based on `_model_CamilleFull` and the CES model by @DaluS (see the next point). Might still need changes.
- changes in nomenclature (remove `-`) to avoid syntax errors. For example, `_model_GK-CES` becomes `model_GKCES`
- `K_0` added to `_def_fields`


DaluS:
That's great ! I implement it manually, with just the modification on `def_fields` and the new model file

DaluS (approved):

PR #158, opened 2022-02-22
Improvement on cycles with harmonics calculation

DaluS:
TobeFinished


DaluS:
@chevaldemetal if you can have a sneak peek that'd be nice !

PR #161, opened 2022-03-03
Climate inequalities

chevaldemetal:
Coping 2018 tested on the three basic scenarios


DaluS:
@guittcam I believe a few files comes from your version and should not be in the library. Can you : 
1. Identify what are the .DStore and remove them (editing the .gitignore might do the trick)
2. Remove _def_fieldsSAVE from the PR
3. either remove `run_model` or put it with a good documentation (comments) in the `doc` folder if you want it to be shared

We need to do such thing to keep a clean code :)

DaluS:
@chevaldemetal the problem is in your units ! For the moment we only consider a limited amount of units acceptable (that can be edited in `def_fields`, with the latex formalism ( units/y is written units.y^{-1} )

Typically executing your model is raising an error : 
image

Let's discuss which units you want to add (I think more than two of them needs new unit in your model to stay consistent), and I recommend executing the last lines of `tutorial.py` to be sure everything works without necessary launching unitary test 


```
listofsolver = pgm.get_available_solvers(returnas=list)
listofsolver = [listofsolver[i] for i in [0]]
listofmodels = pgm.get_available_models(returnas=list)
for model in listofmodels:
    presets = pgm.get_available_models(returnas=dict)[model]['presets']
    for preset in [None]+presets:
        hub = pgm.Hub(model, preset=preset)
        hub.set_dparam(Tmax=1, verb=False)
        for solver in listofsolver:
            print(model, preset, solver)
            hub.run()
```

Or if you want to use pytest directly in the ipython terminal
```
!pytest pygemmes/tests/test_01_Hub.py -v
!pytest pygemmes/tests/test_00_get -v
!pytest pygemmes/tests/test_02_Hub_Multiple -v
!pytest pygemmes/tests/test_03_articles -v
```

DaluS:
@chevaldemetal I added a `Coping2018-VPaul` which is written slightly differently : the values for presets are directly in a dictionary, which makes the file lighter and IMHO more readable. I'm curious about what you think of this writing. 
For the moment I have put comments on all units change to see if the pytests goes through this version :)

The original file has been transfedered in the folder `Hidden` inside `_model` ! That way tests are not runner on it (yet)

DaluS:
Tests not working for the moment, I'll check this Friday (something weird with `time` and a name of file that should not be there)

